PR #6, opened 2023-03-18
fix: ensure correct `updateURL` links

alaidriel:
Seems some of the `updateURL` links were still missed.


sgoudham (approved):
Ah, I didn't catch those. Thanks!

PR #18, opened 2023-03-28
feat: split anichart and anilist (i think)

rubyowo:


rubyowo:
... excuse me
you didn't label it

PR #30, opened 2023-05-05
refactor: gen script

rubyowo:


isabelroses:
@sgoudham lgtm but probably don't just take my word for it

PR #110, opened 2023-07-22
fix(twitch): Fix darkTheme json

Sichelmondprinz:
It gave the error "Invalid JSON: frappe is not a valid JSON literal" because of a missing quotation mark.


isabelroses (approved):

isabelroses:
Thanks for the fix

Issue #69, opened 2023-07-21
Missing or outdated preview images


jn-sena:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### Describe your issue.

The following userstyles don't have the separate previews by flavours, or have outdated combined previews.

- [Cinny](https://github.com/catppuccin/userstyles/tree/main/styles/cinny)
- [Codeberg](https://github.com/catppuccin/userstyles/tree/main/styles/codeberg)
- [DeepL](https://github.com/catppuccin/userstyles/tree/main/styles/deepl)
- [ichi.moe](https://github.com/catppuccin/userstyles/tree/main/styles/ichi.moe)
- [Libreddit](https://github.com/catppuccin/userstyles/tree/main/styles/libreddit)

isabelroses:
Thanks for letting us know, in terms of 
- [Libreddit](https://github.com/catppuccin/userstyles/tree/main/styles/libreddit) - which maybe shutting down
- [Codeberg](https://github.com/catppuccin/userstyles/tree/main/styles/codeberg) - which is getting a rework
For the rest i will take a look (progress tracker below).
- [ ] [Cinny](https://github.com/catppuccin/userstyles/tree/main/styles/cinny)

- [ ] [DeepL](https://github.com/catppuccin/userstyles/tree/main/styles/deepl)
- [ ] [ichi.moe](https://github.com/catppuccin/userstyles/tree/main/styles/ichi.moe)

justTOBBI:
Hey, the Issue on Codeberg is noted. Once I started/finished the rework :)

sgoudham:
I redid the images for `deepl` here: https://github.com/catppuccin/userstyles/commit/573b28a863e252f373462d8b691a15f6a0e914fe

Sorry for not tagging issue!

sgoudham:
Codeberg was the last one listed here, just updated the previews for now and a new set of previews can be added with the upcoming rework.

Thanks everyone for updating these images!

PR #119, opened 2023-07-23
ci: add `file-checker.yml`

sgoudham:
We want to enforce the existence of the files contained within the userstyles template directory. This will reduce manual error when maintainers are reviewing PRs and checking for these files.

**Depends on #69**

### DIFF
--- a/.github/workflows/file-checker.yml
+++ b/.github/workflows/file-checker.yml
@@ -3,9 +3,6 @@ name: Required Files Checker
 on:
   workflow_dispatch:
   pull_request_target:
-  push:
-    branches:
-      - "ci/file-checker" # remove this after testing
 
 jobs:
   file-checker:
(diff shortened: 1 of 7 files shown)


.github/workflows/file-checker.yml
```diff
@@ -0,0 +1,49 @@
...
+
+          for dir in styles/*; do
+            if [ -d "$dir" ]; then
+              find_files "$dir"
+            fi
+          done
+
+          if [ $MISSING_FILES -eq 0 ]; then
+            echo "# No Missing Files 🙌" > $GITHUB_STEP_SUMMARY
+          fi
```
isabelroses, line 46: Other then this being funny, does it serve any purpose since if it does not fail we know all is good?

isabelroses (approved):

.github/workflows/file-checker.yml
sgoudham: This overwrites the previous markdown in the job summary when its a success so that the summary isn't confusing when it passes

PR #125, opened 2023-07-23
fix(vercel): flavor

ryanccn:


nekowinston (approved):
Context: Fixes a previous accidental var rename in #122

PR #130, opened 2023-07-24
ci: add prettier [skip ci]

nekowinston:
Adds prettier, set to only format `./styles/**/*.css*`, with `codeberg` & `reddit` currently excluded, as they are still using the Stylus preprocessor.

@sgoudham and I agreed on formatting on a separate commit rather than just failing a check, requiring the author to format themselves to fix it.


nekowinston:
Merging this now since no CSS PRs are open while writing this, and I really want the commit that changes the formatting to happen on `main`, rather than pinging *everyone* again.

PR #132, opened 2023-07-24
ci: add userstyle linter

nekowinston:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,16 @@
+name: Lint userstyles
```
sgoudham, line 1:
```suggestion
name: Userstyle Linter
```

This matches with the other workflow names, however, we might want to change the others to match with this PR instead.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,16 @@
...
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
```
sgoudham, line 10:
```suggestion
    runs-on: ubuntu-latest
    name: Lint
```

Matching case with the other workflows

.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: Userstyle Linter
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  lint:
+	name: Lint
```
sgoudham, line 10:
```suggestion
    name: Lint
```

src/lint/lint.ts
```diff
@@ -0,0 +1,63 @@
...
+      core.error(err, { file: entry.relative });
+    }
+
+    const assert = assertions(repo);
+
+    Object.entries(assert).forEach(([k, v]) => {
+      const defacto = metadata[k];
+      if (defacto !== v) {
+        const line = css.split("\n").findIndex((line) => line.includes(k)) + 1;
+        core.warning(`Metadata ${k} should be ${v} but is ${defacto}`, {
```
uncenter, line 42: Do warnings support markdown? I would prefer something like this (template values wrapped in inline code backticks, punctuation):
```js
`Metadata key \`${k}\` should be \`${v}\` but is \`${defacto}\`.`
```

> Metadata key `preprocessor` should be `less` but is `stylus`.

uncenter (approved):
LGTM.

PR #133, opened 2023-07-24
fix: correct the typos

isabelroses:
Corrections for all typos show in https://github.com/catppuccin/userstyles/actions/runs/5642097101
~Exclusion for the `tranpart` one from the youtube port.~
I checked the one from deepl and that appears to not be an issue.

@nekowinston, @sgoudham I know you both are interested/invested in this.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -13,6 +13,6 @@ flavours = "flavours"
 
 [type.css]
 extend-ignore-re = [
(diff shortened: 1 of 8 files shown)


nekowinston (approved):
Thanks! I think agreed on not checking class names for now.
This regex should do the trick while still checking CSS property names.

PR #138, opened 2023-07-25
fix(proton): wrong updateURL

soya-daizu:


nekowinston (approved):

nekowinston:
Oops, don't worry about prettier failing, will fix that with the push :sweat_smile:

PR #143, opened 2023-07-26
ci: automatically sync labels with userstyles

### LINKED ISSUE #112: Automate additions/removals to `@userstyles-maintainers`

### Describe your issue.

Currently, adding and removing maintainers within `current-maintainers` in the `userstyles.yaml` does not update `@userstyles-maintainers` team.

This is frustrating because the `CODEOWNERS` file expects that all listed users have write permissions. Without some form of automation, members will have to be manually added and removed which is definitely something we want to avoid as this repository grows in popularity. 

We should utilise GitHub's GraphQL API to automatically update the entire team based on the current state of `current-maintainers` in our CI workflows.

### Acceptance Criteria

- [ ] additions/removals to `current-maintainers` should be reflected in the `@userstyles-maintainers` team

sgoudham:
This PR adds a new CI workflow that will automatically sync the repository labels based on an auto-generated file `.github/labels.yml`, ensuring that maintainers do not have to worry about manually adding/updating or removing labels. 

The `name`, `description` and `color` are captured `userstyles.yml`. If the `color` attribute exists for the userstyle, the relevant color from the macchiato palette will be used, otherwise it will default to `text`. 

I would like to explore a different mechanism for the default label color as most userstyles do not specify a color, possibly generating different shades of the colors? or some hashing based approach. Although, it's probably for the best that it's not over-engineered, we could also default all userstyle labels to be one single colour to avoid bringing in the `@catppuccin/palette` library.

--- 

closes #112


.github/workflows/sync-labels.yml
```diff
@@ -0,0 +1,14 @@
+name: Sync Labels
+
+on:
+  push:
+    branches: [main]
+
```
isabelroses, line 6: Should it not have something like this, so that it only runs when a new port is added?
```suggestion
     paths:
      - .github/labels.yml
```

isabelroses (approved):

PR #151, opened 2023-07-28
docs(proton): prefer Snow base theme on latte

soya-daizu:


nekowinston (changes requested):
I believe the README would have to be changed [here](https://github.com/catppuccin/userstyles/blob/d1f8a425bb4451ff228bfcaa9a9201d489c11ef4/src/userstyles.yml#L267-L274) instead of the Markdown, which is auto-generated.

soya-daizu:
Oops, I didn't know the existence until now! Should I close this PR and then make a new one?

nekowinston:
You can use `git reset --soft HEAD^` to undo your local commit, then `git restore ./styles/` and simply force-push onto your branch.

nekowinston (approved):
Good to merge since the README will get overwritten anyways.

Edit: bonus points for the last commit :)

PR #158, opened 2023-08-03
fix(chatgpt): Add ctp-colors to code blocks

adriankarlen:
I am unsure if I got every type they use, but these were the ones I found.

Added italic and weight to better mimic the look of the ctp vscode theme.


rubyowo:
I think we should just import the https://github.com/catppuccin/highlightjs port

rubyowo:
Superseded by: #186

PR #188, opened 2023-09-03
feat(chatgpt): use highlightjs import

isabelroses:
This uses and `@import` to bring in https://unpkg.com/@catppuccin/highlightjs@0.1.2/css/catppuccin.variables.css
This should allow for easier updates based on the CTP releases
Example:
image


rubyowo (approved):
epic

PR #195, opened 2023-09-13
fix(tutanota): color replacements for most UI elements

ryanccn:
the world if Tutanota used CSS variables


isabelroses (commented):
I would just prefer if we didn't change `@accentColor` but if it unavoidable then its fine

isabelroses (approved):

PR #208, opened 2023-09-19
docs: add version info to general recommendations

adriankarlen:
Add info regarding updating the version number of the userstyle in order for it to be pushed as a new update to users without requiring a reinstall.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -24,6 +24,7 @@ It is important to note that there are **two types of PRs** for this repository:
 - Consider using
   [conventionalcommits.org](https://www.conventionalcommits.org/en/v1.0.0/)'s
   rules for creating explicit and meaningful commit messages.
+- Update the version in the `==UserStyle==` header of the css-file. This is to enable automatic updates for users.
 - If it's your first time contributing to a project then you should look to the
   popular
   [first-contributions](https://github.com/firstcontributions/first-contributions)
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -24,6 +24,7 @@ It is important to note that there are **two types of PRs** for this repository:
 - Consider using
   [conventionalcommits.org](https://www.conventionalcommits.org/en/v1.0.0/)'s
   rules for creating explicit and meaningful commit messages.
+- Update the version in the `==UserStyle==` header of the css-file. This is to enable automatic updates for users.
```
isabelroses, line 27: `This is to enable automatic updates for users.` This isn't exactly true, it would be better if it was rephrased to some form of version control.

isabelroses (changes requested):

docs/CONTRIBUTING.md
adriankarlen: According to https://github.com/openstyles/stylus/issues/39#issuecomment-345953794 while not automatic it makes the style able to be updated via the update url. Guess the phrasing automatic is a bit off, but it still enables updating without reinstall of the style.

isabelroses: I know of it allowing updates it was more about the phrasing of `automatic` rather then the updates.

adriankarlen: Rephrased to: "This is to enable version control of the style."

isabelroses (approved):

PR #217, opened 2023-09-22
Stylus

GitMuslim:
## 🎉 Theme for Stylus 🎉

Stylus is the web extension responsible for every theme in this whole repo

## 💬 Additional Comments 💬

> **Note**
> This theme is still in progress, if you have any comments or find something unthemed, please tell me in this PR! Thus, the catwalk will be generated when the theme is ready for merging.

## 🗒 Checklist 🗒

- [ ] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [x] I have made sure to update the
      [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/src/userstyles.yml)
      file with information about the new userstyle.
  - `catppuccin.user.css` - all the CSS for the userstyle, based on the
    template.
  - `assets/mocha.webp`, `assets/macchiato.webp`, `assets/frappe.webp` and
    `assets/latte.webp` - individual screenshots of the themed website, in all 4
    Catppuccin flavors.
  - `catwalk.webp` - image of all individual screenshots stitched together,
    generated via [catwalk](https://github.com/catppuccin/toolbox#-catwalk).


GitMuslim:
idk if the text editor perfectly matches with the guidelines laws, that's why I didn't check the first checkbox

rubyowo:
you should add an options for accent colors instead of just teal

andreasgrafen:
Doesn't the scope leak styles to outside the stylus extension? Since it's just set to `url-prefix("moz-extension://"), url-prefix(chrome-extension://)` it would technically try to apply the styles to **every** `moz-extension` page, no?

isabelroses:
> Doesn't the scope leak styles to outside the stylus extension? Since it's just set to `url-prefix("moz-extension://"), url-prefix(chrome-extension://)` it would technically try to apply the styles to **every** `moz-extension` page, no?

I think due to some manifest change at some point it shouldn't be able to leak.

andreasgrafen:
> I think due to some manifest change at some point it shouldn't be able to leak, as extensions cannot affect other extensions.

Ah, okay. That's why I was'nt able to test this. But if that's default Firefox and not bc of some hardening it should be fine. Thabks for the clarification. c:

isabelroses (approved):

GitMuslim:
Thanks!

PR #226, opened 2023-09-27
Brave Search – remove searchbar background gradient

andreasgrafen:
I'll assume that this isn't intended as it does look a little weird. It seems like an oversight, or something Brave may have added since this was released:

With Style installed as is there is a weird gradient within the search-bar. That is fixed via this PR.

Before / After
brave search with gradient
brave search without gradient


ndsboy (changes requested):
Hi there, thank you for the PR.
Interesting choice of Brave to add a gradient to the search bar that is not visible when using the default design...

I'd suggest using a css selector without the "svelte-m5b..." selectors, since they look generated and may change.
Otherwise, it looks good to me!

ndsboy (approved):

PR #252, opened 2023-10-04
docs(github): mention GitHub appearance settings

sgoudham:
This commit aims to clearly highlight that the GitHub default theme should be changed via the Appearance settings.

**Are we able to include links in this content?** It would be nice to maybe link to the GitHub docs as that will be updated regularly updated by GitHub themselves.


isabelroses:
A good idea would also usage notes for the theme.

isabelroses (approved):

PR #260, opened 2023-10-08
docs: several error corrections & improved comprehension

uncenter:
This PR adds periods at the ends of sentences in a few places and rephrases some sections to be consistent with the rest.


src/userstyles.yml
```diff
@@ -272,13 +272,13 @@ userstyles:
     readme:
       app-link: "https://github.com/mastodon/mastodon"
       usage: |+
-        Add urls to `@-moz-document domain("url")` with url being the chosen server
+        Add URLs to `@-moz-document domain("url")` with url being the chosen server.
```
uncenter, line 275: This seems poorly phrased but I don't know enough about `@-moz-document domain()` to fix it. Maybe something like "Add Mastodon server URLs to `@-moz-document domain("url")`, e.g. `@-moz-document domain(some example here...)`."

isabelroses: I agree what you mean in terms of phrasing but im unsure what would be better strictly speaking and currently it makes enough sense to get by.
In terms of examples i don't think its needed since the actual user.css contains some examples in context that are a strong hint enough for the user

isabelroses (approved):

README.md
```diff
@@ -58,7 +58,7 @@ To install these userstyles:
 > **Note** <br>
```
uncenter, line 58: Are we intentionally disabling the GitHub callout syntax here or?

src/userstyles.yml
uncenter: What if we use the examples from the `user.css` in the README?

> Add your Mastodon server URLs to `@-moz-document domain("url")`, like this:
> ```
> @-moz-document domain("mstdn.social"),
> domain("fosstodon.org"),
> domain("tech.lgbt"),
> domain("mastodon.social")
> ```

README.md
isabelroses: What do you mean?

uncenter: Screenshot 2023-10-08 at 13 53 37 (Arc)

vs.

> [!NOTE]
> If you would like to submit a userstyle or learn how to port Catppuccin to a website, please refer to the [userstyle-creation.md](docs/userstyle-creation.md).

uncenter: What we have right now looks like the old GitHub callout syntax and I was guessing the `<br>` was there to disable it?

isabelroses: This can be swapped then, this was just an oversight and not spotted before then

uncenter: Alright. Should I do that  -  switch all of the `> **Note** <br>`s to `> [!NOTE]`s (in `userstyles.yml` and `./README.md`) - in this PR?

isabelroses: Yes go ahead

src/userstyles.yml
uncenter: Update on this @isabelroses?

isabelroses: I mean that could make sense but only 2 or so would be more than enough to show how it is done.

uncenter: How about this?

Add your Mastodon server URLs at the top of the userstyle, like this:
```
@-moz-document domain("mastodon.social"), domain("fosstodon.org") {
```

isabelroses: Sounds good to me

uncenter: Or maybe we change this part of the `user.css` to this:

```diff
- @-moz-document domain("mstdn.social"),
- domain("fosstodon.org"),
- domain("tech.lgbt"),
- domain("mastodon.social"),
- domain("piaille.fr") {
+  @-moz-document domain("mastodon.social") {
```

And have the README say:

Replace `@-moz-document domain("mastodon.social") {` at the top of the userstyle with the Mastodon server URLs that you use, like this:

```css
@-moz-document domain("mastodon.social"), domain("fosstodon.org") {
```

isabelroses: I think, this makes the user experience worse, i would prefer to have 100+ default urls then to lose any for the point of an example.
But we can certainly put a comment notice in to tell the use to do something similar for their own instance

uncenter: I think we should aim to be as user friendly as possible and make sure we are careful about how we guide people who may not be comfortable with using/editing userstyles. Saying to "add your Mastodon server URLs at the top of the userstyle" might be confusing about where exactly the should add it, if they can remove the other URLs there, etc. I also think comments in the userstyle are much less likely to be read than if you put a note in the README.

Also another thought just occurred to me - won't editing the URLs be removed when you update the userstyle? Should there be a note about that?

isabelroses: Yeah, the searxng port has a similar issue where you can have multiple urls, they have all the default urls, and users can add their own instance if they wish.

What we are aiming for here is something similar.  Where a user will find it simple to add their own instances if they need, but ideally they would not have to due to how many instances we support by default.

uncenter: I'm thinking...

---

This userstyle is applied by default on a few common Mastodon server URLs, listed at the top:
```css
@-moz-document domain("mastodon.social"), domain("mstdn.social"), domain("fosstodon.org"), domain("tech.lgbt") {
```
If you don't see your Mastodon server URL, feel free to add it to the list with the `domain("<URL>")` syntax.

---

And what do you think about my other comment?
> won't editing the URLs be removed when you update the userstyle? Should there be a note about that?

isabelroses: I love the new note suggestion i think it captures the situation perfectly.

> won't editing the URLs be removed when you update the userstyle? Should there be a note about that?

In terms of this yes it will. I think that:
- We can prompt users to either contribute their instance to the userstyle domain pool
- re add their instance upon update

uncenter: Alright. I'll just put the clarified note in the README and maybe you can add the two bullet points you suggested in a separate PR?

isabelroses: Sure that makes sense

uncenter:
Good to go I think!

uncenter:
Sorry little nitpick lol.

isabelroses (approved):
Thanks for the hard work and suffering i put you though.

PR #262, opened 2023-10-08
docs: update browsers.md

uncenter:
This PR updates `docs/browsers.md` to use the new GitHub callout syntax, `[!NOTE]` (sorry I missed this file earlier!) and corrects the syntax for the browsers table.

### DIFF
--- a/docs/browsers.md
+++ b/docs/browsers.md
@@ -3,7 +3,7 @@
 > [!NOTE]
 > If you see a popular browser that hasn't been added, please create a pull request to do so!
 
-Find your browser version by following the instructions below:
+Find your browser version by following the link for your browser.
 
 <table>
   <thead>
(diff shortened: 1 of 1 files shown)


isabelroses (approved):

uncenter:
Is it worth adding `arc://version/` for Arc in this PR actually?

isabelroses:
> Is it worth adding `arc://version/` for Arc in this PR actually?

If Arc is compatible with the stylus extension and its not Chromium based etc, i don't see why not

uncenter:
> > Is it worth adding `arc://version/` for Arc in this PR actually?
> 
> If Arc is compatible with the stylus extension and its not Chromium based etc, i don't see why not

I'm using Arc right now! It is Chromium-based but some people... might not know that.

isabelroses:
> I'm using Arc with Stylus right now! It is Chromium-based but... some people might not know that.

Im happy for it to be added, then

isabelroses (approved):

PR #263, opened 2023-10-10
ci: bump actions and make titles lowercase

uncenter:
I'm not sure why each step's name is title case - doesn't really fit with GitHub actions casing (see below), though I do understand title case workflow titles, happy to change that back. Otherwise, this PR bumps all uses of `actions/checkout` to `v4` and `stefanzweifel/git-auto-commit-action` to `v5`. I also removed the `commit_author` field used with `stefanzweifel/git-auto-commit-action` since it is already the default with the action.

Screenshot 2023-10-09 at 21 40 08 (Arc)

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -1,4 +1,4 @@
-name: File Generator
+name: File generator
 
 on:
   workflow_dispatch:
@@ -13,27 +13,24 @@ env:
 jobs:
   generate-health-files:
     runs-on: ubuntu-latest
-    name: Generate Health Files
+    name: Generate health files
     permissions:
       # Give the default GITHUB_TOKEN write permission to commit and push the
       # added or changed files to the repository.
       contents: write
 
     steps:
-      - name: Check Out Repository
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: env.HAS_USERSTYLES_TOKEN
         with:
           token: ${{ secrets.USERSTYLES_TOKEN }}
           ref: ${{ github.ref }}
-
-      - name: Check Out Repository (Fork)
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: env.HAS_USERSTYLES_TOKEN == 'false'
         with:
           ref: ${{ github.ref }}
 
-      - name: Set up Deno
+      - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
           deno-version: v1.x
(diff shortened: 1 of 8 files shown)


.github/workflows/generate.yml
```diff
@@ -13,27 +13,24 @@ env:
...
     runs-on: ubuntu-latest
-    name: Generate Health Files
+    name: Generate health files
     permissions:
       # Give the default GITHUB_TOKEN write permission to commit and push the
       # added or changed files to the repository.
       contents: write
 
     steps:
-      - name: Check Out Repository
```
uncenter, line 23: Rationale: none of the other workflows have a name for the `actions/checkout` step.

.github/workflows/generate.yml
```diff
@@ -13,27 +13,24 @@ env:
...
           ref: ${{ github.ref }}
-
-      - name: Check Out Repository (Fork)
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: env.HAS_USERSTYLES_TOKEN == 'false'
         with:
           ref: ${{ github.ref }}
 
-      - name: Set up Deno
```
uncenter, line 36: "Set up" vs "setup" is arguable, I have no issue reverting back to "set up".

.github/workflows/typos.yml
```diff
@@ -1,28 +1,28 @@
...
     # 'continue-on-error' will always succeed, we should exit based on the outcome and not the conclusion
     # ref: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context
-    - name: Fail When Typos Are Found in Pull Request
+    - name: Fail when typos are found in pull request
       if: ${{ github.event_name == 'pull_request' && steps.typos.outcome != 'success' }}
       run: "exit 1"
 
     # We always want this check to pass on branch 'main' but continue to fail on PRs
-    - name: Pass When Branch 'main'
+    - name: Pass on 'main' branch
```
uncenter, line 26: I think this flows better than "Pass when branch 'main'".

.github/workflows/generate.yml
```diff
@@ -44,15 +41,14 @@ jobs:
...
         run: ./generate.ts
 
-      - name: Push Changes
-        uses: stefanzweifel/git-auto-commit-action@v4
+      - name: Push changes
+        uses: stefanzweifel/git-auto-commit-action@v5
         if: ${{ github.repository == 'catppuccin/userstyles' && github.ref == 'refs/heads/main' && env.HAS_USERSTYLES_TOKEN }}
         with:
           commit_message: "chore: generate health files"
-          commit_author: "github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>"
```
sgoudham, line 56: I believe this was added because the default string will include the person who merged the PR as a co-author, and that doesn't really make sense. 

In the words of @nekowinston 
> there's no reason i'm co-authored when i merge a PR in that changes a single line in the readme :baldkek:

Please revert this line :+1:

uncenter: image
That explains this!

uncenter: Y'know... [the other action I suggested](https://github.com/EndBug/add-and-commit) doesn't have this problem :D
But yeah for consistency and all that I'll revert it aha.

uncenter:
Also - thoughts on dependabot for updating actions, once a month maybe?

sgoudham (changes requested):
Thanks for going ahead and updating the actions.

In terms of the reasoning of title case, I personally think it's an easy to differentiate built in steps versus your own steps but I don't really mind either way. However, I do think having the action names as title case is nicer so that it stands out more in the checks.

With regards to dependabot, I'm not that much of a fan of it. I have it enabled monthly for another repository and its sometimes a pain to deal with. I think we can keep on top of the actions without needing it :+1: 

If you wouldn't mind, could you please revert the changes just for the `name:` keys and we should be good to merge. Thanks!

uncenter:
Totally! Will do in a minute. And thanks for your thoughts on Dependabot, that's what I was thinking but wanted to make sure.

sgoudham (approved):

PR #266, opened 2023-10-13
fix(stylus): light+dark mode

GitMuslim:
The user should be able to change themes from stylus settings
image

the theme had the light/dark mode thingy connected to system settings rather than stylus, this PR fixes it

as for "By system preference", this is automatically changed by stylus to either dark or light depending on the system theme


isabelroses (commented):
Can you  disgard commits that are redundant, alot of this PR is just replacing things that were already there, making it harder to observer the actual changes

Issue #274, opened 2023-10-21
Scroll bar too for youtube userstyle. Request for a setting to set the width in pixel


Crystal4276:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:youtube

### Describe your problem.

The size of the scroll bar is too small on my resolution. I'm having difficulty to grab it.
With this old version there was the possibility to set the width in pixel of the scroll bar.
Can we add it back ?
```/* ==UserStyle==
@name YouTube Catppuccin-Mocha
@namespace github.com/catppuccin/YouTube
@homepageURL https://github.com/catppuccin/YouTube
@version 1.3.5```
```
### Attach screenshots.

![image](https://github.com/catppuccin/userstyles/assets/27446196/02632b0f-869e-4640-b7c7-6ff7ccc04a59)

### What browser(s) are you seeing the problem on?

thorium

### Any additional comments?

Thanks guys for your amazing work.

PR #276, opened 2023-10-22
feat(yt): improve scroll bar

isabelroses:
Closes https://github.com/catppuccin/userstyles/issues/274


Crystal4276:
Hello,

Just tested, there is two issues:
- The scrollbar width parameter doesn't seem to have an effect,
- There is one of the two scroll bar which doesn't have the cattppuccin theme applied

image

isabelroses:
> * The scrollbar width parameter doesn't seem to have an effect,
> * There is one of the two scroll bar which doesn't have the cattppuccin theme applied. see the right scroll bar on the screenshot

Thats strange since mine comes out perfectly:
image
image

Can you tell me what your scrollbar input reads?

PR #307, opened 2023-11-14
fix(yt): various fixes

jopejoe1:
The notification bell and bell next to sub button also theme YouTube's feature banner


isabelroses (approved):
Thanks for the PR

PR #310, opened 2023-11-14
Fix Verified Badge Coloring

GenShibe:
before: 
image
after: image

### DIFF
--- a/styles/instagram/catppuccin.user.css
+++ b/styles/instagram/catppuccin.user.css
@@ -440,16 +440,6 @@
       @mauve 60%,
       @blue 90%
     );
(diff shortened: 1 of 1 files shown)


isabelroses (changes requested):
Can you bump the version number.
And can we remove the redundant code from https://github.com/catppuccin/userstyles/pull/310/files#diff-a4f6ca944c84e122ae7cb64d252d6b9fc24d862468b8a5a5a70fa14bf2e59885L441 and a bit beyond that.

GenShibe:
i THINK i've addressed all those, unless there's other places beyond the CSS file that the version number needs to be bumped up

isabelroses (approved):
Thanks gen.

PR #312, opened 2023-11-14
chore(template): update

nekowinston:
@isabelroses some of those changes we discussed.


isabelroses (approved):
Thanks for doing this.

PR #316, opened 2023-11-14
chore: add VSCode settings

nekowinston:
Working on this repo with VSCode requires some special setup & extensions, why not use the built in feature to share those settings - should be less of a hassle for everyone.


uncenter (approved):
Good idea!

PR #317, opened 2023-11-15
feat: init stylelint

uncenter:
Definitely a WIP. Adds linting with Stylelint!

## Resources

- https://stylelint.io/user-guide/rules
- https://lesscss.org/functions/

## Todo

- [x] Configure general rules
- [ ] Add Stylelint CI workflow

PR #318, opened 2023-11-15
ci(lint): init

nekowinston:
This finishes what https://github.com/catppuccin/userstyles/pull/132 started, supersedes https://github.com/catppuccin/userstyles/pull/317 (thanks for the stylelint setup!) so it runs in a single step.


uncenter:
Haven't had a chance to review but thanks for putting this together! Looks great from what I can tell.

PR #346, opened 2023-11-20
feat(lichess): init

coopw1:
## 🎉 Theme for [lichess.org](https://lichess.org/)🎉
'Lichess is a free and open-source Internet chess server run by a non-profit organization of the same name.' 

mocha
wooo
## 💬 Additional Comments 💬
Forever ago, but was requested [here](https://github.com/catppuccin/catppuccin/discussions/1831) 

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [x] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [x] I have ensured that the new directory is in **lower-kebab-case**.
  - [x] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have made sure to update the
      [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/src/userstyles.yml)
      file with information about the new userstyle.
- [x] I have included the following files:
  - `catppuccin.user.css` - all the CSS for the userstyle, based on the
    template.
  - `assets/mocha.webp`, `assets/macchiato.webp`, `assets/frappe.webp` and
    `assets/latte.webp` - individual screenshots of the themed website, in all 4
    Catppuccin flavors.
  - `catwalk.webp` - image of all individual screenshots stitched together,
    generated via [catwalk](https://github.com/catppuccin/toolbox#-catwalk).

PR #350, opened 2023-11-21
build(lint): use native stylelint config resolution

uncenter:
Prevent errors on custom elements in #346, see https://github.com/catppuccin/userstyles/pull/346#issuecomment-1820213474.


nekowinston:
I really liked the added typo protection that `selector-type-no-unknown` added, and I think it's rare enough where a workaround for each userstyle isn't that much effort. I'll push what I came up with.

nekowinston:
Lichess lint passes for me when I create `./styles/lichess/.stylelintrc.json` with:
```json
{
  "rules": {
    "selector-type-no-unknown": [
      true,
      { "ignore": ["signal", "index", "fill"] }
    ]
  }
}
```

nekowinston:
[8ed27aa](https://github.com/catppuccin/userstyles/pull/350/commits/8ed27aa4dbc29190a38b7b72dfa9dd984eae3702) works but this really isn't getting any more maintainable.
Luckily it turned out that turning on [`nodeModulesDir`](https://docs.deno.com/runtime/manual/getting_started/configuration_file#nodemodulesdir) and not passing in `config` into `stylelint.lint` does everything we want for us.

The config file will have to include an `extends` to the repo root, compared to the previous impl:
```json
{
  "extends": "../../.stylelintrc.js",
  "rules": {
    "selector-type-no-unknown": [true, { "ignore": ["signal", "index", "fill"] }]
  }
}
```

nekowinston (approved):
Think this should do it.

uncenter:
That's way better, thanks!

PR #381, opened 2023-12-10
build: fix path of schema & update commit hash

sgoudham:
## 🔧 What does this fix? 🔧

The latest PR on `catppuccin/catppuccin` (https://github.com/catppuccin/catppuccin/pull/2222) changed the schema and refactored the paths. In turn, this has broken the generation script. 

This PR updates the commit hash of the schema and updates the path

## 🗒 Checklist 🗒

~~- [ ] I have updated the version appropriately in the `==UserStyle==` header of the `catppuccin.user.css` file.~~


deno.lock
sgoudham, line 1: This doesn't seem right... did I do something wrong here?

nekowinston: Yeah, I think only a single `*.ts` file is being added to the cache, I usually do `deno cache **/*.ts --lock-write`.

nekowinston (approved):

PR #382, opened 2023-12-10
docs: add `looking for maintainers` section

sgoudham:
## ✨  Experimenting with the Schema ✨

This PR tries to introduce the concept of _**unmaintained userstyles**_ into the repository, with the hopes of starting some conversation as I'd like to try and apply this to the wider organisation (of course, after a maintainer audit...)

I've picked on @rubyowo here because I remember her saying that she doesn't really want to maintain any userstyles (if you didn't, I'm sorry but you're getting picked on anyways.) Will we hope and pray that Ruby changes her mind? Will we keep relying on @isabelroses to try and work on userstyles that they might not necessarily want to work on? Not necessarily saying that's the case but some food for thought :P 

In terms of the changes, I've

- Updated the handlebars template to display a looking for maintainers message if no `current-maintainers` array is present.
- Modified the schema to not require `current-maintainers` anymore.
- Modified the code to handle the `current-maintainers` array being `undefined` 

By no means are these changes elegant or even should be accepted, but I think it's time we started properly trying to work around the fact that people in open source lose motivation, move on, deal with other things, etc. (properly trying = trying to document this in metadata and display the information on READMEs)

Appreciate any thoughts.

## 🗒 Checklist 🗒

~~- [ ] I have updated the version appropriately in the `==UserStyle==` header of the `catppuccin.user.css` file.~~


rubyowo:
The AOC port is pretty stable, and i wouldn't mind maintaining it (because there's nothing to do)

sgoudham:
> The AOC port is pretty stable, and i wouldn't mind maintaining it (because there's nothing to do)

Yeah AOC isn't a great example there but I was just trying to get at the fact that maintainership essentially comes down "who cares if an issue gets raised, a PR is raised, the website changes, etc"

It's not even about getting fixes out in a timely fashion or adding features (as you said, things can be stable and that's fine) but it's just about having someone that cares and can eventually conjure up the motivation to work on it right. 

The chatgpt port is probably a good example here since they might change the UI tomorrow and break a bunch of stuff, at that point I'd expect the maintainer listed against the userstyles to _care_ right?

rubyowo:
Also i would recommend adding a "Picking up userstyles" section to CONTRIBUTING.md and linking to it from the readme

isabelroses:
> Will we keep relying on @isabelroses to try and work on userstyles that they might not necessarily want to work on? Not necessarily saying that's the case but some food for thought :P

I don't mind working on unmaintained styles because its rare that they need large updates, however it would completely unsustainable, so I completely support this PR.

scripts/generate/main.ts
```diff
@@ -56,6 +56,7 @@ await updateFile(
   join(REPO_ROOT, ".github/CODEOWNERS"),
   Object.entries(userstylesData.userstyles)
     .map(([slug, { readme }]) => {
+      if (readme["current-maintainers"] === undefined) return "";
```
nekowinston, line 59:
```suggestion
      if (!readme["current-maintainers"]) return;
```

nekowinston (approved):
Just one minor style suggestion, rest LGTM.

We need to remember to bump the catppuccin/catppuccin schema, since this makes changes to `userstyles.readme > required []` which would invalidate the data against the schema over on the main repo.

scripts/generate/main.ts
sgoudham: I was actually going to maybe suggest keeping `current-maintainers` required but allowing for an empty list? I like idea of a **visible** empty list as you're scrolling through the list.

Also speaking to @rubyowo on discord, I think it would only be the chatgpt port that could be regarded as "unmaintained" right now?

nekowinston: > I was actually going to maybe suggest keeping `current-maintainers` required but allowing for an empty list? I like idea of a **visible** empty list as you're scrolling through the file.

Sure, I can see the benefits of that.

sgoudham:
Hello everyone that's been requested for a review :wave: 

This PR assumes that you are no longer willing to maintain your userstyle and therefore is removing you, based on the userstyle not being updated with regards to issues raised. I realise that this may seem sudden and quite harsh. 

I will aim to publish a document that precisely defines what maintainership means for Catppuccin :+1:

AnubisNekhet:
Pretty much all of my userstyles are practically complete, i.e. they don't really need any further work unless there is a breaking change or an unthemed area I might have missed. I'm still available for those. There hasn't been any particular issue raised on any my styles as far as I can tell and I have not experienced any theming issues on those specific sites. TLDR; The userstyles are essentially complete (for now) though I'm still available for resolving any issues.

isabelroses (approved):

styles/chatgpt/README.md
uncenter, line 1: Why are these being edited in this PR? Shouldn't this be done by the CI?

PR #399, opened 2023-12-13
feat(import-styles): init

isabelroses:
- Allows for a mass import of userstyles.
- We can enforce settings upon users making our lives easier.

PR #402, opened 2023-12-14
ci(lint): validate `name` metadata

uncenter:


scripts/lint/metadata.ts
```diff
@@ -2,17 +2,20 @@
...
 import { log } from "@/lint/logger.ts";
 import * as color from "std/fmt/colors.ts";
+import * as path from "std/path/mod.ts";
 import { sprintf } from "std/fmt/printf.ts";
 import type { WalkEntry } from "std/fs/walk.ts";
 import { relative } from "std/path/mod.ts";
-import { REPO_ROOT } from "@/deps.ts";
+import { REPO_ROOT, userStylesSchema } from "@/deps.ts";
+import { UserstylesSchema } from "@/types/userstyles.d.ts";
+import { validateYaml } from "@/utils.ts";
```
uncenter, line 11: Not sure how y'all want me to sort these imports...

isabelroses:
This should probably also validate other `@`'s in the userstyles meta block too. For example using `@license` outside will also create an error.

uncenter:
I think you misunderstand. This lints the actual value of `@name`. I think what you are suggesting is to make sure there aren't other `@name`/`@<usercss-metadata-variable>` outside of the meta block?

uncenter:
Screenshot 2023-12-14 at 07 17 05 (Arc)

Other than the first two it is working as expected. Verifies that the name provided in the `userstyles.yml` file matches the actual userstyle's name. `WhatsApp Web` has an extra space, others have capitalization issues or differ in some way.

isabelroses:
> I think you misunderstand. This lints the actual value of `@name`. I think what you are suggesting is to make sure there aren't other `@name`/`@<usercss-metadata-variable>` outside of the meta block?

ohhh that makes sense, thanks for informing me

uncenter:
> > I think you misunderstand. This lints the actual value of `@name`. I think what you are suggesting is to make sure there aren't other `@name`/`@<usercss-metadata-variable>` outside of the meta block?
> 
> ohhh that makes sense, thanks for informing me

Np. Not a bad idea though, but this PR (and the linter in general) will in a way let us know if `@name` (or other vars) is/are used elsewhere since that will be flagged when parsed incorrectly.

scripts/lint/metadata.ts
```diff
@@ -67,9 +70,23 @@ export const verifyMetadata = (
...
+const assertions = async (repo: string) => {
   const prefix = "https://github.com/catppuccin/userstyles";
+
+  const { userstyles } = await validateYaml<UserstylesSchema>(
+    Deno.readTextFileSync(path.join(REPO_ROOT, "scripts/userstyles.yml")),
+    userStylesSchema,
+  );
+  if (userstyles === undefined) {
+    Deno.exit(1);
+  }
```
uncenter, line 82: Adapted from https://github.com/catppuccin/userstyles/blob/main/scripts/sync-maintainers/main.ts#L14-L20. Should we just make this code a function like `getUserstylesData` in `@/utils.ts`?

uncenter:
Semi-dependent on #399 btw.

nekowinston:
> Semi-dependent on #399 btw.

in the sense of it throwing warnings until that PR is merged?

uncenter:
Yep, just that there is an issue with the usercss-meta package where if you have `@name` (or any `@<variable>` from the meta block) anywhere outside of the meta block, it messes up the parsing. We fixed it in #399 since it was relevant there.

nekowinston (approved):
LGTM then, don't really care about the warnings showing up for the next few commits.

uncenter:
There are some legitimate issues raised by the addition of this PR though. Should I fix them here or in a separate PR after this merges?

nekowinston:
Sure, let's go for the mass ping again 😅

styles/lichess/catppuccin.user.css
```diff
@@ -1,5 +1,5 @@
 /* ==UserStyle==
-@name Lichess Catppuccin
+@name lichess Catppuccin
```
uncenter, line 2: I think we should make the name uppercase in `userstyles.yml` (and here), see https://github.com/catppuccin/userstyles/pull/346#discussion_r1398709738 for the conversation from a while back

styles/wikiwand/catppuccin.user.css
```diff
@@ -1,5 +1,5 @@
 /* ==UserStyle==
-@name Wikiwand Catppuccin
```
uncenter, line 2: Actually we need to change `userstyles.yml` here as well:
Screenshot 2023-12-14 at 11 46 00 (Arc)

uncenter:
Bruh-

uncenter:
@nekowinston is there a reason we are using `esm.sh` for type-fest but the `npm:` modifier for other NPM packages?

nekowinston:
Yeah, type-fest is already in cache since our deno-stdlib uses ESM imports.

uncenter:
Ohhh right neat.

PR #404, opened 2023-12-15
fix(ci/lint-build): add GH_TOKEN to enviroment

isabelroses:
fix for failing ci https://github.com/catppuccin/userstyles/actions/workflows/lint-build.yml


nekowinston (approved):

nekowinston (changes requested):
oh nevermind actually i don't know if the extra token we provided has those permissions

PR #418, opened 2023-12-27
fix: missed a few lint annotations

uncenter:


isabelroses (approved):

styles/deepl/catppuccin.user.css
```diff
@@ -182,10 +182,6 @@
     color: darken(@accent-color, 15%);
   }
 
-  [class*="bg-[#B4DAE8]"] {
-    // background-color: fadeout(@accent-color, 70%);
-  }
-
```
uncenter, line 188: Not sure if this was a note-to-self or something. There's another reference to the same selector on line 194/195 with another commented out `background-color` rule as well.

styles/deepl/catppuccin.user.css
```diff
@@ -591,10 +587,6 @@
     }
   }
 
-  [class|="Quote-module--quote"] {
-    border-color: @surface0;
-  }
-
```
uncenter, line 597: `border-color` is already set for the same selector on line 584.

PR #419, opened 2023-12-27
ci(lint): exit with error when issues are found

uncenter:
@nekowinston

### DIFF
--- a/scripts/lint/stylelint.ts
+++ b/scripts/lint/stylelint.ts
@@ -13,9 +13,15 @@ import { log } from "@/lint/logger.ts";
 export const lint = async (entry: WalkEntry, content: string, fix: boolean) => {
   const file = relative(REPO_ROOT, entry.path);
 
+  let success = true;
+
   await stylelint.lint({ files: entry.path, fix })
     .then(({ results }) => {
       results.map((result) => {
+        if (result.warnings.length > 0) {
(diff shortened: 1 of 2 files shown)


uncenter:
The CI failing means it is working :)

nekowinston:
Thanks, just changed some of my old bad code there :)

nekowinston (approved):

scripts/lint/stylelint.ts
```diff
@@ -24,14 +22,14 @@ export const lint = async (entry: WalkEntry, content: string, fix: boolean) => {
...
-            file,
+            file: relative(REPO_ROOT, entry.path),
             startLine: warning.line,
             endLine: warning.endLine,
             startColumn: warning.column,
             endColumn: warning.endColumn,
             content,
           }, warning.severity);
         });
+        if (result.warnings.length > 0) throw new Error("stylelint error");
```
uncenter, line 33: Why would we through here? Wouldn't it not get through all of the files?

uncenter: @nekowinston 👀

nekowinston: This is invoked over in
```ts
for await (const entry of stylesheets) {
  lint()
```
and we only throw after all the warnings have been printed.

PR #420, opened 2023-12-28
ci(scripts): consistent variables & comments

uncenter:


uncenter:
#420! @nekowinston any chance you could take a look?

isabelroses (approved):

scripts/lint/metadata.ts
```diff
@@ -32,32 +31,29 @@ export const verifyMetadata = async (
...
 
-      const message = sprintf(
-        "Metadata %s should be %s but is %s",
-        color.bold(k),
-        color.green(v),
-        color.red(String(defacto)),
-      );
+      const message = `Metadata ${color.bold(key)} should be ${
+        color.green(value)
+      } but is ${color.red(String(defacto))}`;
```
sgoudham, line 43: This looks harder to read than the existing code, I think the existing code is fine here as it is.

sgoudham (changes requested):
One comment from me, looks okay otherwise

scripts/lint/metadata.ts
uncenter: Hmm. I think it's more readable the other way, but I can revert. Do you want me to revert all such changes of `sprintf`?

sgoudham: I think this block is the only place with `sprintf`? 

I think this is less readable since prettier is splitting across multiple lines in an unnatural way, would appreciate if you can revert.

Basically just leave the `const message = ...` as it was before.

sgoudham (approved):

PR #423, opened 2023-12-29
fix(github): regex

### LINKED ISSUE #422: Github userstyle regex fails to match certain users and orgs

lbl:github

### Describe your problem.

The regex for github userstyle fails to match users and organizations whose name starts with `home`, `marketplace`, etc. I discovered this on https://github.com/homebridge, any repo under their org is not themed.

image
As you can see, the regex used to exclude `/home` also excludes `homebridge`

### Attach screenshots.

image

Microsoft Edge 120.0.2210.91

### Any additional comments?

_No response_

nekowinston:
Closes #422

https://regexr.com/7pqa8


BlankParticle (changes requested):
Fixes the issue I mentioned,
when we are at rewriting the regex we should also exclude https://github.com/features

BlankParticle (approved):
LGTM, I think we can merge it

nekowinston:
I'd prefer being explicit over composition with `\/foo|\/bar` vs `\/(foo|bar)`, the 2nd gets super messy

isabelroses:
> I'd prefer being explicit over composition with `\/foo|\/bar` vs `\/(foo|bar)`, the 2nd gets super messy

This makes sense to me.

PR #432, opened 2024-01-09
fix(github): purge unmaintainable mess

isabelroses:
So in order to make github more maintainable, I decided to purge the absolute mess we have in our userstyle. In order to prevent regressions this is going to need a lot of work.

Largely caused due to the mess that is github's beta features as so many people are on their own different set of betas.


styles/github/catppuccin.user.css
```diff
@@ -13,6 +13,9 @@
 @var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]
 @var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
 @var select accentColor "Accent" ["rosewater:Rosewater*", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+
+@var checkbox search-beta "beta: search" 1
+@var checkbox repo-beta "beta: search" 1
```
uncenter, line 18:
```suggestion
@var checkbox search-beta "[BETA]: Enable Search" 1
@var checkbox repo-beta "[BETA]: Enable Repo" 1
```

Something in the format of that would be better, and we should actually make clear what those features are. What is "search"? Is that like the command palette? What is "repo"? Is that like the new code view, or the repository overview refresh?

isabelroses: That's a good point, i should clarify it more.

styles/github/catppuccin.user.css
```diff
@@ -13,6 +13,9 @@
 @var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]
 @var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
 @var select accentColor "Accent" ["rosewater:Rosewater*", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+
+@var checkbox search-beta "[BETA]: Enable Search" 1
+@var checkbox repo-beta "[BETA]: Enable New Repo Display (i.e. file browser)" 1
```
uncenter, line 18:
```suggestion
@var checkbox repo-beta "[BETA]: Enable New Repository Code View" 1
```

uncenter (changes requested):

PR #474, opened 2024-01-28
fix(modrinth): update branding color vars

izmystic:
fixes the release tag not being green

## 🔧 What does this fix? 🔧

Makes it so the release tag on projects stays green

## 🗒 Checklist 🗒


izmystic:
before:
image
after:
image

uncenter:
The scope here should be of the userstyle. So `fix(modrinth): update branding color vars`.

PR #478, opened 2024-01-28
fix(bsky): some unthemed parts 

RoootTheFox:
bluesky changed some inlined colors that the userstyle matches for, so things broke

there's probably some things that I missed, i did this as a really quick fix, will probably do a proper check later today

### DIFF
--- a/styles/bsky/catppuccin.user.css
+++ b/styles/bsky/catppuccin.user.css
@@ -55,7 +55,7 @@
 
     /* they dont even use this, get real */
     --text: @text;
-    --background: @mantle;
+    --background: @base; /* they do use this, kind of! */
     --backgroundLight: @base;
 
     /* generic text */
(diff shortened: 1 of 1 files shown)


styles/bsky/catppuccin.user.css
```diff
@@ -64,16 +64,19 @@
...
       color: @accent-color !important;
     }
 
     /* blue svg inline */
+    [fill="hsl(211, 99%, 53%)"] {
+      fill: @accent-color;
+    }
     [fill="#0085ff"] {
       fill: @accent-color;
     }
```
uncenter, line 82:
```suggestion
    /* blue svg inline */
    [fill="hsl(211, 99%, 53%)"],
    [fill="#0085ff"] {
      fill: @accent-color;
    }
```

styles/bsky/catppuccin.user.css
```diff
@@ -255,7 +258,7 @@
     path[fill="#ffffff"] {
       fill: @text;
     }
-    path[fill="#fff"] {
+    path[fill="hsl(211, 20%, 100%)"] {
       fill: @text;
     }
```
uncenter, line 263: It won't let me make a suggestion here but:
```less
    path[fill="#ffffff"] {
      fill: @text;
    }
    path[fill="hsl(211, 20%, 100%)"] {
      fill: @text;
    }
```
could be condensed to:
```less
    path[fill="#ffffff"],
    path[fill="hsl(211, 20%, 100%)"] {
      fill: @text;
    }
```

uncenter (changes requested):

styles/bsky/catppuccin.user.css
RoootTheFox: oh yea, i didn't have the compact syntax in my original theme and was a bit too tired to be bothered to refactor. I'll do the changes you mentioned when I get home later today; as commit signing on my iPad is currently broken (?)

isabelroses (approved):

PR #482, opened 2024-01-30
fix(lastfm): Optimize masks

AnubisNekhet:


isabelroses (changes requested):
just need to bump version number and your good to go.

isabelroses (approved):

PR #485, opened 2024-01-31
ci: consolidate format & generate workflows

uncenter:
Fixes the race condition between the two workflows committing to the repository by combining them into one. Requires switching the commit workflow from `stefanzweifel/git-auto-commit-action` to `EndBug/add-and-commit` because the former lacks a way to disable pushing the commit immediately.


sgoudham (approved):
Looks fine, guess we won't really know until it's live.

sgoudham (changes requested):
Actually, just caught something. We can remove the if conditions on the add and commit action and move it to the git push step? Since that's what we want to prevent when it's running on a fork.

uncenter:
Right. Shouldn't we move it to be an if statement for the whole job? Just don't run any of it? Why would we do that only on individual steps?

uncenter:
Ah, turns out we don't have the `env` context in a` jobs.<id>.if`. Shoot.

sgoudham (approved):

PR #488, opened 2024-01-31
ci: run prettier with deno instead of npx

uncenter:


nekowinston:
We might want to add `npm:prettier` into the lockfile then, so that my action keeps it in cache.
You could create a stub `_foo.ts` with
```ts
import "npm:prettier";
```
to get it to stick to cache.

uncenter:
Like that?

PR #522, opened 2024-02-16
ci(lint): add checks for variables

uncenter:
This is some atrocious code and probably isn't worth adding this mess but I thought I'd give linting variables a try.


uncenter:
You can go to the files tab and scroll all the way to the bottom to see the annotations this adds.

scripts/lint/logger.ts
```diff
@@ -56,7 +56,7 @@ const prettyPrint = (
         color.dim(String(startLine + 1)),
         color.dim(lines[startLine]),
       ),
-      sprintf("%*s╰─► %s", pad, "", error),
+      sprintf("%*s╰─► %s", pad, "", error.split('\n').map((str, i) => i === 0 ? str : " ".repeat(pad + 4) + str).join('\n')),
```
uncenter, line 59: This change pads multiline error messages to have the same indentation as the first line.

scripts/lint/metadata.ts
```diff
@@ -8,6 +8,7 @@ import { relative } from "std/path/mod.ts";
 import { REPO_ROOT } from "@/deps.ts";
 import { log } from "@/lint/logger.ts";
 import { formatListOfItems, getUserstylesData } from "@/utils.ts";
+import stringify from "npm:json-oneline-stringify";
```
nekowinston, line 11:
```suggestion

const stringify = (v: unknown) => JSON.stringify(v, null, 0);
```

nekowinston (commented):
We can replace this with JSON.stringify AFAIK, output on my end still looked fine

scripts/lint/metadata.ts
uncenter: I mean it's less readable... for example it doesn't space apart arrays or objects.
image

nekowinston: don't think we should care for this lint output

uncenter: I think we should because making these warnings readable is pretty key.

Screenshot 2024-02-17 at 06 29 03 (Arc)

uncenter:
My primary concern is that this output won't be understandable since it doesn't match what you see for each variable in the CSS file (i.e. not the `@var select ...` syntax) since it is the parsed version. Thoughts?

```
Found mismatch in array elements of property "options" of metadata variable `lightFlavor`:
+ Expected: { "name": "frappe", "label": "Frappé", "value": "frappe" }
- Received: { "name": "frappe", "label": "Frappe", "value": "frappe" }
```

nekowinston:
what about just saying
```
variable `lightFlavor` should be `[ "latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha" ]`
```
so you actually have something to copy; it'd be a few hardcoded strings but only for light/dark flavors, or flavor, + accent.

Also we should probably default to mauve instead of sapphire since that is what all other UI stuff of big ports defaults to.

uncenter:
> what about just saying
> 
> ```
> variable `lightFlavor` should be `[ "latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha" ]`
> ```
> 
> so you actually have something to copy; it'd be a few hardcoded strings but only for light/dark flavors, or flavor, + accent.

Hmm, I've given this some thought and take a look at the lint annotations now. I think they are very readable, however I didn't  make it "copyable" because often it was very hard to tell what was actually wrong with what you had.

> Also we should probably default to mauve instead of sapphire since that is what all other UI stuff of big ports defaults to.

This would probably be a separate follow up PR since I'm just going off of the template but yeah sounds good.

nekowinston:
@uncenter 
> Hmm, I've given this some thought and take a look at the lint annotations now. I think they are very readable, however I didn't make it "copyable" because often it was very hard to tell what was actually wrong with what you had.

I just meant if the assertions against `lightFlavor`/`darkFlavor` (or `flavor` for non-syncing themes), and `accent` fail, we know exactly what kind of string it should say; which is why I said
> it'd be a few hardcoded strings

uncenter:
Yeah I implemented that and it didn't look right. It seems like being able to copy it would be the easiest but it was frustrating because it was difficult to figure out *why* what you had was wrong.

nekowinston:
It doesn't matter why its wrong, we literally know what string should be there from the template

uncenter:
In my opinion understanding why the linter is yelling at you is pretty important.

nullishamy:
For the boilerplate header, the "why" isn't actually important. It's the same for every single userstyle. For any non-experienced contributors (who are the real targets / likely recipients of these lints), it's important to be concise and precise. With a JSON log, you must manually inspect it, interpret what it means, and apply those changes. This is difficult.

There is little correlation between the JSON output and the original Stylus header, which makes understanding / fixing the problem at hand significantly harder, and will overall increase friction between potential contributors and mantainers. 

On the topic of interpretation, without colour the diff is extremely hard to read. This is exacerbated by the fact that this lint will often fail for very subtle copy errors (as you showed in the example).

With the problem source of this lint (copy errors) in mind, it only makes sense to directly inform readers that this is the case.
Here is the message structure that I think effectively communicates this idea:
```
Invalid boilerplate detected (did you copy it wrong?).
Variable `lightFlavor` should be `@var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]`
```
Lints are for contributors, not maintainers, so we should cater to them.

uncenter:
> For the boilerplate header, the "why" isn't actually important. It's the same for every single userstyle. For any non-experienced contributors (who are the real targets / likely recipients of these lints), it's important to be concise and precise. With a JSON log, you must manually inspect it, interpret what it means, and apply those changes. This is difficult.
> 
> There is little correlation between the JSON output and the original Stylus header, which makes understanding / fixing the problem at hand significantly harder, and will overall increase friction between potential contributors and mantainers.
> 
> On the topic of interpretation, without colour the diff is extremely hard to read. This is exacerbated by the fact that this lint will often fail for very subtle copy errors (as you showed in the example).

If you look at the current lint annotations you will see that the JSON output is no longer in use.

> With the problem source of this lint (copy errors) in mind, it only makes sense to directly inform readers that this is the case. Here is the message structure that I think effectively communicates this idea:
> 
> ```
> Invalid boilerplate detected (did you copy it wrong?).
> Variable `lightFlavor` should be `@var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]`
> ```
> 
> Lints are for contributors, not maintainers, so we should cater to them.

From what I can gather it seems like people prefer to know what they did wrong though. I expect a contributor might compare what the linter suggests to what they have, and it is usually a very subtle difference and frustrating as they don't understand what they had wrong.

As of right now, the lint annotations look like this:
```
Metadata variable `accentColor`: default should be "sapphire" but is "blue"
```
```
Metadata variable `lightFlavor`: label of option `frappe` should be "Frappé" but is "Frappe"
```

Maybe a combination of "suggest what they should have" and "tell them precisely what the issue is" would be best?

```
Metadata variable `lightFlavor`: label of option `frappe` should be "Frappé" but is "Frappe" and ...
Use `@var select lightFlavor ...` instead.
```

nullishamy:
I don't see why we need to be this specific. The contributor does not need to care what is actually wrong, it's a boilerplate. If it were lints in code they wrote, absolutely we should be specific. This is not the case, however.

As for your latter suggestion, I'm not sure why we'd duplicate the information? Above details why they don't need specifics, just the general case.

uncenter:
> I don't see why we need to be this specific. The contributor does not need to care what is actually wrong, it's a boilerplate. If it were lints in code they wrote, absolutely we should be specific. This is not the case, however.
> 
> As for your latter suggestion, I'm not sure why we'd duplicate the information? Above details why they don't need specifics, just the general case.

Alright. I've made the changes, please check the lint annotations now and tell me what you think :)

scripts/lint/metadata.ts
```diff
@@ -20,35 +20,81 @@ export const verifyMetadata = async (
...
+
+    if (current === 0) {
+      // This variable is undefined so there isn't a line for it, so we just put it at the bottom of the variables section.
+      const line = lines
+        .findLastIndex((line: string) => line.includes("==/UserStyle== */")) +
+        1;
+
+      log(
+        sprintf(
+          "Metadata variable `%s` should exist",
```
nullishamy, line 78: Can we suggest what the value should be here?

nullishamy:
> > I don't see why we need to be this specific. The contributor does not need to care what is actually wrong, it's a boilerplate. If it were lints in code they wrote, absolutely we should be specific. This is not the case, however.
> > As for your latter suggestion, I'm not sure why we'd duplicate the information? Above details why they don't need specifics, just the general case.
> 
> Alright. I've made the changes, please check the lint annotations now and tell me what you think :)

Looks good now I think, just one question there!

scripts/lint/metadata.ts
uncenter: cc: @sgoudham

isabelroses (approved):

scripts/lint/metadata.ts
sgoudham: Yeah it causes visual clutter since there is a GitHub user called `@var` and you're not able to escape it in the lints? and I believe there's currently no workaround.

I'm not sure if it pings them but I don't think we'd want to do it anyways, I think the current log message is _fine_. We can revisit it at a later point if the lint ends up causing too much confusion

nullishamy: Alright, cool, seems fine. If you blunder the copy this bad, that's on you 👀 .

nullishamy (approved):
We can revisit refactoring this to be nicer at some point in the future, but it works, and isn't terribly big, so not a huge deal.

sgoudham:
> So now the question is how we resolve the lint errors. Separate follow up PR? Some of the errors are about userstyles that only have a single flavor (instead of both darkFlavor and lightFlavor)... should be an easy fix with media queries?

2 questions:

1. Just to confirm my knowledge, they're not technically `errors` and just `warnings` and wouldn't cause the PR check to fail right? 
2. Is there a _valid_ case here where a userstyle wouldn't want to have both variables set? Or is it the case that its always set?

sgoudham (approved):
A PR to fix the warnings seems like a good next step after merging this 👍

PR #529, opened 2024-02-17
fix: correct spellings of flavor

isabelroses:
Based off of the information learnt from https://github.com/catppuccin/userstyles/pull/522 we still use some British spellings of flavor which we don't want to do.


rubyowo:
Boston Tea Party

uncenter (approved):

PR #574, opened 2024-02-28
fix(duckduckgo): region switcher button

### LINKED ISSUE #570: duckduckgo language switch should not be colored if disabled

lbl:duckduckgo

### Describe your problem.

Duckduckgo has a switch right on top to select if the search results should be in a specific language. In the current version this switch is always colored in the accent color, no matter if it is disabled or enabled. This makes it hard to tell in which state the switch is.

#### With this userstyle:

disabled
enabled

#### In vanilla duckduckgo:

disabled
enabled

Firefox 122.0.1

### Any additional comments?

I'd propose to set the color to `surface0`.

(I'm also willing to work on this myself, but I don't have the time right now and maybe someone already acquainted with contributing here is faster than me)

GenShibe:
## 🔧 What does this fix? 🔧
closes #570 

## 🗒 Checklist 🗒

### DIFF
--- a/styles/duckduckgo/catppuccin.user.css
+++ b/styles/duckduckgo/catppuccin.user.css
@@ -510,10 +510,13 @@
       color: @crust !important;
     }
 
-    .js-region-filter-switch,
+ .js-region-filter-switch,
     .frm__switch__label {
-      background-color: @accent-color !important;
+      background-color: @overlay0;
       border-color: @accent-color !important;
+    }
+    .switch.is-on {
+      background-color: @accent-color;
     }
 
     .dropdown__switch.is-on::before,


isabelroses (approved):

styles/duckduckgo/catppuccin.user.css
```diff
@@ -510,11 +510,14 @@
...
 
-    .js-region-filter-switch,
+ .js-region-filter-switch,
     .frm__switch__label {
-      background-color: @accent-color !important;
+      background-color: @overlay0;
       border-color: @accent-color !important;
     }
+    .switch.is-on {
+      background-color: @accent-color;
```
atticus-sullivan, line 519: Seems like `!important` is needed here so that the color really is being applied (now the color (some blue) from duckduckgo isn't changed when the switch is active)

Issue #601, opened 2024-03-02
Syntax highlighting colors for Wikipedia.


aspizu:
### Is there an existing issue outlining your problem?

- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:wikipedia

### Describe your problem.

Syntax highlighting colors for Wikipedia.

### Attach screenshots.

Need to add css something like:

```css
.mw-highlight .nf {
    color: #89b4fa;
}

.mw-highlight .kt {
    color: #f38ba8;
}

.mw-highlight .c1, .mw-highlight .cm {
    color: #6c7086;
}

.mw-highlight .mi {
    color: #74c7ec;
}

.mw-highlight .k {
    color: #a6e3a1;
}

.mw-highlight .nb {
    color: #94e2d5;
}

.mw-highlight .mf, .mw-highlight .mh, .mw-highlight .mb {
    color: #9399b2;
}

.mw-highlight .cp {
    color: #fab387;
}

.mw-highlight .cpf {
    color: #eba0ac;
}

.mw-highlight .s {
    color: #f9e2af;
}

.mw-highlight .se {
    color: #fab387;
}
```

### What browser(s) are you seeing the problem on?

Firefox v112.0.2

uncenter:
Can you provide a link to a page with syntax highlighting? Makes it easier to fix if I don't have to go out and think of a page that has that lol.

aspizu:
It uses pygments, we should create a pygments style and use that.

aspizu:
<https://github.com/wikimedia/mediawiki-extensions-SyntaxHighlight_GeSHi/blob/master/modules/pygments.generated.css>

uncenter:
That seems to use `pygments` - I wonder if we can generate it with https://github.com/catppuccin/python/blob/465e3269e943d2ae492069f7ecedb6ce9799e0fa/catppuccin/extras/pygments.py#L2?

aspizu:
```py
from catppuccin.extras.pygments import MochaStyle
from pygments.formatters.html import HtmlFormatter
css = HtmlFormatter(style=MochaStyle).get_style_defs()
open("mocha_pygments.css", "w").write(css)
```

uncenter:
[...]

PR #605, opened 2024-03-03
feat(wikipedia): Theme syntax highlighting, Fix line numbers in code-blocks, and stylistic changes. (#601)

aspizu:
I removed the border from code-blocks as I believe it looks better without it.

### DIFF
--- a/styles/wikipedia/catppuccin.user.css
+++ b/styles/wikipedia/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name Wikipedia Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/wikipedia
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/wikipedia
-@version 0.0.2
+@version 0.0.3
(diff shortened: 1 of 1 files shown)


isabelroses (approved):
LGTM

styles/wikipedia/catppuccin.user.css
```diff
@@ -256,10 +256,266 @@
 
     .mw-highlight pre {
       background: @surface0 !important;
-      border-color: @surface2 !important;
+      border-width: 0;
```
isabelroses, line 259: Oh wait sorry, just saw this. Is this avoidable?

aspizu: Sure, can be undone if you wish.

aspizu:
```css
border-color: @surface0 !important;
```

aspizu: But, it makes the border look like this:

isabelroses: Alright, that makes sense.

PR #610, opened 2024-03-04
chore(tree-wide): remove non-catwalk images

uncenter:
As discussed in Discord, hopefully will fix our Git LFS bandwidth issue.


BlankParticle (approved):

justTOBBI (approved):
Lgtm!

sgoudham:
To add extra context to this PR. 

We're starting to run out of git-lfs bandwidth from the assets on this repository:
image

I briefly looked into the "data packs" that GitHub offer to increase this limit, and it looks like you can only increase storage and bandwidth **at the same time**. One data pack is 5$ per month and would bump us to 50GB storage and 50GB bandwidth.

While we do have the money to fund this, I'm not comfortable with spending OC money on this and feel like it would be a waste of money (compared to purchasing domains, mail server, etc)

@isabelroses raised a good point that most people, probably, dont bother viewing the individual image previews so removing them here makes the most sense right now. I suppose we'll have this discussion again if/when the bandwidth runs out from just the `catwalk` images.

uncenter:
> I suppose we'll have this discussion again if/when the bandwidth runs out from just the `catwalk` images.

I think we'd have to get something like 3 times as much traffic as we do now to pass the 1GB bandwidth threshold with only the `preview.webp` images.

Tnixc:
Is there a reason the images must be hosted on github? There's plenty of free image hosts out there. Not sure if we discussed this before.

uncenter:
> Is there a reason the images must be hosted on github?

Hosting in GitHub makes it easier to maintain and contribute to IMO.

> There's plenty of free image hosts out there.

Any suggestions/ideas/examples of those?

sgoudham:
> Is there a reason the images must be hosted on github? There's plenty of free image hosts out there. Not sure if we discussed this before.

Personally, I followed the template at the beginning without necessarily thinking about the scale in which this repository might grow into. I think just having the catwalk image is fine instead of needing the 4 individual image previews for each folder in this repository (and/or needlessly investing into an image hosting service)

Tnixc:
Yeah that makes sense

isabelroses (commented):
Why not flatten the directories a bit by having the preview in the same dir as the readme and theme.

isabelroses (changes requested):
Please also update the template.

TheMemeSniper (approved):

Tnixc (approved):

uncenter:
> Please also update the template.

What do you mean "the template"?

uncenter:
> Why not flatten the directories a bit by having the preview in the same dir as the readme and theme.

Isn't it assets/preview.webp in catppuccin/template or similar?

RoootTheFox (approved):

isabelroses:
> > Please also update the template.
> 
> What do you mean "the template"?

Currently the template also has the old previews too. i.e. mocha.webp

isabelroses:
> > Why not flatten the directories a bit by having the preview in the same dir as the readme and theme.
> 
> Isn't it assets/preview.webp in catppuccin/template or similar? I guess we could? I'll leave it up to you/Hammy.

It is like that in catppuccin/template. But since we are moving away from that template here, I don't see why we can't flatten the directories out.

sgoudham:
> hy not flatten the directories a bit by having the preview in the same dir as the readme and theme.

I like this, since `assets/preview.webp` doesn't really give any further meaning than just having a `preview.webp` in the same directory as the other files.

uncenter:
> > > Please also update the template.
> > 
> > 
> > What do you mean "the template"?
> 
> Currently the [template](https://github.com/catppuccin/userstyles/tree/main/template/assets?rgh-link-date=2024-03-04T13%3A35%3A01Z) also has the old previews too. i.e. mocha.webp

Ah I didn't even realize the template had preview images... what are they even screenshots of lol??

isabelroses (approved):

PR #614, opened 2024-03-04
ci(pr-labeler): disable on tree-wide PRs

uncenter:
Disables it from running on PRs with a title containing `(tree-wide)` since those often edit so many files that the labeler workflow fails due to too many labels.

Side note: looks like we are pinned at a very specific version for the `actions/labeler` workflow - any reason for this? Can we / should we upgrade to https://github.com/actions/labeler/releases/tag/v5.0.0?


isabelroses (approved):

isabelroses:
We should probably also document this.

PR #621, opened 2024-03-05
ci(typos): ignore .user.css files

uncenter:
I think it is best to solve the immediate problem of the CI failing on every PR for unrelated reasons. We can come back to this when there is a better solution?


uncenter:
I'll do a follow up PR to update the Chess.com generation script (I believe new versions of catppuccin/python use en-US spellings) and we can remove that as well.

isabelroses (approved):

PR #642, opened 2024-03-08
ci(lint-build): ignore `docs/` folder

isabelroses:
Goals:
- Save some gha run minutes

### DIFF
--- a/.github/workflows/generate-format.yml
+++ b/.github/workflows/generate-format.yml
@@ -4,6 +4,10 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
+    paths:
+      - "styles/**/*"
+      - "scripts/generate/*.ts"
+      - "scripts/userstyles.yml"
 
 env:
   HAS_USERSTYLES_TOKEN: ${{ secrets.USERSTYLES_TOKEN != '' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/generate-format.yml
```diff
@@ -4,6 +4,10 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
+    paths:
+      - "styles/**/*"
+      - "scripts/generate/*.ts"
```
uncenter, line 9: This should be any file in the `scripts/generate` directory right? This doesn't include the templates, for instance.

.github/workflows/lint-build.yml
```diff
@@ -3,8 +3,14 @@ name: Lint & Build
...
   workflow_dispatch:
   pull_request:
+    paths:
+      - "**/*.user.css"
+      - "**/*.ts"
   push:
     branches: [main]
+    paths:
+      - "**/*.user.css"
+      - "**/*.ts"
```
uncenter, line 13: When the `userstyles.yml` file changes this should run as well? IIRC it lints that too?

uncenter (changes requested):

uncenter (approved):

sgoudham (approved):
I thought more about it and I realised that what I wanted was to just ignore the docs folder for the lint-build action.

I think its going to get annoying in the future if we keep adding specific paths

PR #603, opened 2024-03-03
feat(tabnews): init

Guaxinim5573:
## 🎉 Theme for [TabNews](https://tabnews.com.br) 🎉

[TabNews.com.br](https://tabnews.com.br) is a website focused around the community of the Brazil's tech field, aimed at debates and exchange of knowledge through publications and comments created by the users themselves.

<details><summary>Latte</summary></details>
<details><summary>Frappe</summary></details>
<details><summary>Macchiato</summary></details>
<details><summary>Mocha</summary></details>

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [x] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [x] I have ensured that the new directory is in **lower-kebab-case**.
  - [x] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have made sure to update the
      [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml)
      file with information about the new userstyle.
- [x] I have included the following files:
  - `catppuccin.user.css` - all the CSS for the userstyle, based on the
    template.
  - `assets/mocha.webp`, `assets/macchiato.webp`, `assets/frappe.webp` and
    `assets/latte.webp` - individual screenshots of the themed website, in all 4
    Catppuccin flavors.
  - `catwalk.webp` - image of all individual screenshots stitched together,
    generated via [Catwalk](https://github.com/catppuccin/toolbox/tree/main/catwalk#readme).

PR #664, opened 2024-03-12
chore(tabnews): assorted updates

Guaxinim5573:
Continuation of #603 and some fixes.


isabelroses:
Hey, thanks for the updated PR with the changes made from before, we currently have removed all the previews, and moved the assets/catwalk.webp to preview.webp.

isabelroses (commented):
Can you update the version number, then you should be good to go.

isabelroses (approved):

PR #741, opened 2024-04-02
fix(github): danger buttons, notifications bug

uncenter:
## 🔧 What does this fix? 🔧

Themes danger buttons properly, fixes a bug with unread notifications on latte.

## 🗒 Checklist 🗒


isabelroses:
I think that following gh itself, the default icon color should be text and on hover the background color as accent should be less faded.

isabelroses (approved):

PR #771, opened 2024-04-06
fix(youtube): hide weird canvas behind video player

uncenter:
## 🔧 What does this fix? 🔧

| Unthemed | Before | After |
| --- | --- | --- |
| image | image | image

## 🗒 Checklist 🗒


isabelroses:
Thats not something that needs to be fixed, that is the applications ambient mode. Which has an in app toggle. We apply
`#cinamatic { mix-blend-mode: lighten; }` so that it is still themed unless the user wants to remove that "canvas" via the toggle.

isabelroses:
Also we can improve the ambient mode by removing 
https://github.com/catppuccin/userstyles/blob/11389caaad0daa08906d205fca778b2520f1c5d1/styles/youtube/catppuccin.user.css#L396

PR #774, opened 2024-04-07
docs(README): link to `userstylesyml.md` for maintainer requests

sgoudham:
I don't think you can do relative markdown links with anchors in other files, happy to change it if you can.

I've followed the link format widely seen in [GitHub's documentation](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository#troubleshooting-cloning-errors).

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,10 @@ See [CONTRIBUTING.md](docs/CONTRIBUTING.md).
 ## 🖌 Userstyles
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


README.md
uncenter, line 1: Looks good. Do we need to introduce these forced line breaks though?

sgoudham: Found it was easier to view/edit it. I think it's a non issue unless we're enforcing line length but I'd like to avoid that.

uncenter (approved):

PR #775, opened 2024-04-07
refactor(youtube): rewrite

uncenter:
## 🔧 What does this fix? 🔧

Rewrites the userstyle, 50% less lines, less crud :). Variables are kept the same for the most part, though we should try to prune those a bit. I got rid of the extra mobile styling for now but I will come back to it if needed.

## 🗒 Checklist 🗒

### DIFF
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -449,6 +449,11 @@
       --ytd-author-comment-badge-icon-color: @text !important;
     }
 
+    /* Ambient mode */
+    #cinematics {
+      mix-blend-mode: lighten;
+    }
+
     .ytp-settings-button.ytp-hd-quality-badge::after,
     .ytp-settings-button.ytp-hdr-quality-badge::after,
     .ytp-settings-button.ytp-4k-quality-badge::after,
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -58,11 +58,11 @@
     @white: if(@lookup = latte, @base, @text);
     @black: if(@lookup = latte, @text, @base);
 
-    color-scheme: if(@lookup = latte, light, dark);
     color: @text;
+    color-scheme: if(@lookup = latte, light, dark);
 
     ::selection {
-      background: fade(@accent-color, 30%);
+      background-color: fade(@accent-color, 30%);
     }
 
     input,
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -441,6 +441,14 @@
       --yt-button-payment-background-color: @accent-color !important;
     }
 
+    ytd-author-comment-badge-renderer {
+      --yt-basic-background-color: @surface0 !important;
+      --yt-basic-foreground-title-color: @surface0 !important;
+      --ytd-author-comment-badge-background-color: @surface0 !important;
+      --ytd-author-comment-badge-name-color: @text !important;
+      --ytd-author-comment-badge-icon-color: @text !important;
+    }
+
     .ytp-settings-button.ytp-hd-quality-badge::after,
     .ytp-settings-button.ytp-hdr-quality-badge::after,
     .ytp-settings-button.ytp-4k-quality-badge::after,
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -599,7 +599,7 @@
 
     /* Video description */
 
-    #snippet [style="color: rgb(255, 255, 255);"] {
+    [style="color: rgb(255, 255, 255);"] {
       color: @text !important;
     }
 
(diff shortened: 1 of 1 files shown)


isabelroses:
Yeah mobile is technically its own site with its own respective styling, so the theme just doesn't work there anymore.

styles/youtube/catppuccin.user.css
```diff
@@ -425,1067 +442,324 @@
...
-    }
+    .ytp-menuitem {
+      &:not([aria-disabled="true"]):hover {
+        background-color: @surface1;
+      }
 
-    /* ambient mode */
-    #cinematics {
-      mix-blend-mode: lighten;
-    }
```
isabelroses, line 803: this should probably be added back so we have ambient mode support

styles/youtube/catppuccin.user.css
```diff
@@ -425,1067 +442,324 @@
...
-      }
-      /* liked */
-      > g:nth-child(1),
-      > g:nth-child(3) {
-        > path:nth-child(1) {
-          fill: @accent-color !important;
-        }
-        > path:nth-child(2) {
-          stroke: @accent-color !important;
-        }
```
isabelroses, line 680: I personally liked this as it added accent-color likes.

styles/youtube/catppuccin.user.css
```diff
@@ -425,1067 +442,324 @@
...
     }
 
-    /* creator badges */
-    ytd-author-comment-badge-renderer {
-      --yt-basic-background-color: @surface0 !important;
-      --yt-basic-foreground-title-color: @surface0 !important;
-      --ytd-author-comment-badge-background-color: @surface0 !important;
-      --ytd-author-comment-badge-name-color: @text !important;
-      --ytd-author-comment-badge-icon-color: @text !important;
-    }
```
isabelroses, line 798: removing this breaks 
image

styles/youtube/catppuccin.user.css
```diff
@@ -390,7 +410,6 @@
       background-color: @base !important;
     }
 
-    /* transparent header bar */
     #container.ytd-masthead {
       --iron-icon-fill-color: @text !important;
       background-color: @base !important;
```
isabelroses, line 415: Already themed by vars
```suggestion
```

uncenter:
> Yeah mobile is technically its own site with its own respective styling, so the theme just doesn't work there anymore.

It doesn't work with my refactor or just at all before this PR?

isabelroses:
> > Yeah mobile is technically its own site with its own respective styling, so the theme just doesn't work there anymore.
> 
> It doesn't work with my refactor or just at all before this PR?

It breaks with this PR.

styles/youtube/catppuccin.user.css
isabelroses: Actually if you merge/rebase with main this change will already have been made.

styles/youtube/catppuccin.user.css
```diff
@@ -424,925 +441,353 @@
...
-      &:hover {
-        background: lighten(@accent-color, 10%) !important;
-      }
-    }
-
-    .html5-ypc-description,
-    .html5-ypc-title,
-    .html5-ypc-action-heading {
-      color: @text;
+    ytd-author-comment-badge-renderer {
```
isabelroses, line 444: This should fix some other small issues caused by this.
```suggestion
    ytd-author-comment-badge-renderer:not([style*="--ytd-author-comment-badge-icon-background-color: transparent;]) {
```

styles/youtube/catppuccin.user.css
```diff
@@ -424,925 +441,353 @@
...
-    /* menu items */
-    /* checkbox subitem */
-    .ytp-menuitem-toggle-checkbox {
-      background-color: @overlay1;
-
-      &::after {
-        background-color: @subtext0;
-      }
+    [fill="white"] {
+      fill: if(@lookup = latte, @base, @text) !important;
```
isabelroses, line 476: This seems to break the live indicator, since its bg is `@accent-color` it becomes unreadable on dark themes.

isabelroses (changes requested):
Expanded video descriptions change to white text.

Outside of these comments LGTM

isabelroses:
Just spotted per channel links seem broken too
image

styles/youtube/catppuccin.user.css
uncenter: I think I've fixed it if I understood you correctly. Can you check now?

uncenter:
> Just spotted per channel links seem broken too image

Fixed! Thanks, I had meant to check channel pages but it slipped my mind

uncenter:
> Expanded video descriptions change to white text.

Fixed as well. Mistakenly used a more specific selector to only select that text when it was shortened/in snippet form. Good catch!

isabelroses (commented):
Looks like the subscribe button broke somewhere along the way
image

uncenter:
> Looks like the subscribe button broke somewhere along the way image

In what way is it broken? Looks perfectly fine to me.

isabelroses (approved):

PR #811, opened 2024-04-12
refactor: hoist `current/past-maintainers` keys up

sgoudham:
This PR hoists the `current-maintainers` and `past-maintainers` keys up, outside of the `readme`. I didn't put much thought about it being under `readme` at the time since it didn't really matter but this change makes it more consistent with the `ports.yml` in catppuccin/catppuccin.

The main benefit of this change is so that it's _somewhat_ less verbose to retrieve the maintainers from both userstyles and ports if you aren't validating via the schema.

Draft until I update the docs. (I also did this while having dead time on holiday and will be away the next few days so happy for others to push to this PR before I come back to it)


uncenter:
Could also be worth moving `readme.app_link` to `url` to match the ports.yml schema?

sgoudham (commented):
Thanks for updating the docs.

I think the `url` change sounds fine, best for another PR if anything.

uncenter:
Sounds good, I'll do it in a follow up PR. If my docs commit was enough for you this is good to merge I think!

uncenter (approved):

PR #815, opened 2024-04-13
fix(youtube): theming for skeleton

CD-Z:
## 🔧 What does this fix? 🔧

This PR adds Theming to the skeleton when loading the youtube homepage.

image
Note: Top is new opened in chromium and bottom is old opened in firefox

## 🗒 Checklist 🗒

### DIFF
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -455,11 +455,7 @@
     #home-chips,
     #search-container {
       background-color: @base;
-    }
-
-    .masthead-skeleton-icon,
-    #menu-icon {
-      z-index: 1;
+      z-index: -1;
     }
 
     #guide-skeleton .guide-ghost-icon,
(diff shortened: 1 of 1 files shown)


styles/youtube/catppuccin.user.css
```diff
@@ -449,6 +449,26 @@
...
+    #home-container-skeleton,
+    #home-chips,
+    #search-container {
+      background-color: @base;
+    }
+
+    .masthead-skeleton-icon,
+    #menu-icon {
+      z-index: 1;
+    }
```
isabelroses, line 463: Is this needed?

isabelroses (changes requested):

styles/youtube/catppuccin.user.css
CD-Z: Yes, for some reason these two menu items disappear when the ```background-color``` of ```#search-container``` gets changed.
I already tried moving the css blocks and add ```!important``` but I only got it to work with this workaround.

isabelroses (approved):

PR #830, opened 2024-04-21
fix(github): update markdown headings zindex

ndom91:
Update **GitHub** Userstyle

- Add style for `markdown-headings`, i.e. `#`/`##` in markdown bodys (like `README.md`'s).

## Before

image

## After

image

### DIFF
--- a/styles/github/catppuccin.user.css
+++ b/styles/github/catppuccin.user.css
@@ -512,6 +512,11 @@
       content: url("https://giscus.catppuccin.com/assets/loading_48x48.gif");
     }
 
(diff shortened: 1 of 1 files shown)


isabelroses:
Hello there, I apricate the fix. But this is an issue with github themselves and we would rather not alter the default website since if at some point github decides to fix or change this in any way then we would need to change this too. And as a general rule with catppuccin userstyles is that we try to refrain from "opinionated" changes like these where we change the look and feel of the website rather then just the colors.

ndom91:
Okay makes sense, thanks for the quick turnaround :sweat_smile:

PR #832, opened 2024-04-22
feat(wikipedia): add option for redirect highlight

hf02:
## 🔧 What does this fix? 🔧

This puts the green redirect highlight behind an option, as it does not seem to be Wikipedia's default behavior and feels opinionated.

I've done this instead of outright requesting to remove it, as the [contributing guide](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md) does not seem to outright prevent this since it's just a color change. However, it feels against the spirit of the opinionated section to me, and I also just don't like the green 🙃

On (previously the only behavior):
image

Off (added in this PR):
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/wikipedia/catppuccin.user.css
+++ b/styles/wikipedia/catppuccin.user.css
@@ -351,7 +351,6 @@
     .mw-redirect when (@highlight-redirect = 1) {
         color: @green !important;
     }
-    
 
     .mbox-text {
       color: @text !important;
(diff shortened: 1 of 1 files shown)


isabelroses:
> This puts the green redirect highlight behind an option, as it does not seem to be Wikipedia's default behavior and feels opinionated.

I actually quite like this and the implementation you went with works pretty well. Thank you.

isabelroses (approved):

PR #833, opened 2024-04-22
fix(youtube): low contrast for video player progress

hf02:
## 🔧 What does this fix? 🔧

Low contrast for the YouTube video player's progress bar

Before, the background was unstyled and could be hard to see at times on Mocha.

image

After, the background now uses a translucent Surface 0, and the buffer indicator uses an opaque Overlay 0, making it easier to see:

image
image
image

The white background makes the buffer indicator slightly harder to see, but I do not think this is an issue as the YouTube default also has bad contrast:

image

Latte's contrast suffers from this, so the progress bar is untouched there.

image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -423,10 +423,10 @@
     }
     
     .ytp-progress-list {
-      background: fade(@surface1, 30%);
+      background: fade(@surface0, 80%);
     }
     .ytp-load-progress {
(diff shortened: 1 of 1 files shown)


isabelroses (approved):
LGTM, just want @uncenter to confirm since they are more up to date with the YouTube style

uncenter (approved):

PR #835, opened 2024-04-23
ci(website): dont run on forks

Lichthagel:
Prevents deploy from running in forks

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    if: github.repository == 'catppuccin/userstyles'
     environment: website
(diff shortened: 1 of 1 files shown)


sgoudham (changes requested):
Two things:

1. I'm not really sure if this check is needed since secrets aren't passed to forks so the action would end up failing. I do agree that it's a nice thing to add if someone is hard forking userstyles and keeps merging to their own main.

2. The better way to implement this check would be at the job level and checking the `github.repository` key, as implemented in https://github.com/catppuccin/nix/blob/cc89781d13834e0207ec6fb50ccb2d5122b274c3/.github/workflows/update-docs.yml#L11-L16

Happy to approve once the check is changed.

sgoudham (approved):

PR #949, opened 2024-06-04
feat(mdbook): appply accent-color to links and sidebar text

AxerTheAxe:
## 🔧 What does this fix? 🔧

As things are now, the accent color is almost never used throughout MDbook. This fixes that by changing the link and active sidebar text to use the accent-color instead of always appearing blue.

## 🗒 Checklist 🗒

### DIFF
--- a/styles/mdbook/catppuccin.user.css
+++ b/styles/mdbook/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name mdBook Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/mdbook
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/mdbook
-@version 0.0.2
+@version 0.0.3
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/mdbook/catppuccin.user.css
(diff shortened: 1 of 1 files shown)


uncenter:
Thanks for making this PR! As the mdBook userstyle is based on the upstream https://github.com/catppuccin/mdbook port, you'll need to make a PR with these changes there first.

AxerTheAxe:
Thanks for the info! I will create a upstream PR shortly.

AxerTheAxe:
Upstream gets it's colors from https://github.com/catppuccin/palette which as far as I can tell does not have any way of defining an accent color. Is there any way around this?

uncenter:
> Upstream gets it's colors from [catppuccin/palette](https://github.com/catppuccin/palette) which as far as I can tell does not have any way of defining an accent color. Is there any way around this?

Ah, you are right. I think we can't really do this one, and links anyway are technically supposed to blue/lavender...

AxerTheAxe:
Okay. Thanks for taking time to review this!

uncenter:
Sorry again!

Issue #930, opened 2024-05-30
drive: unstyled svgs


trinkey:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:google-drive

### Describe your problem.

The icons on the upload panel are unstyled

### Attach screenshots.

![image](https://github.com/catppuccin/userstyles/assets/97406176/6441069a-4b2a-4d93-a54d-1b582ef9fc8d)

### What browser(s) are you seeing the problem on?

firefox stable latest

GenShibe:
<img width="402" alt="image" src="https://github.com/catppuccin/userstyles/assets/29279972/818fffef-3923-46f9-9fbe-49a670ef33f3">

it seems like that the icons on the left are *not* svgs, rather PNGs, but the green checkmark *is* an svg

trinkey:
great consistency skills from google

PR #956, opened 2024-06-06
ci(issue-labeler): don't add impartial matches

uncenter:
Fixes an issue with the issue-labeler regex that considers impartial matches - for example, https://github.com/catppuccin/userstyles/issues/930 was given both `google` and `google-drive` due to our regex:

Screenshot 2024-06-06 at 09 13 49 (CleanShot)

Using `^` and `$` should only match full patterns, though I can't really test this.


uncenter:
> I see the idea, my only concern is issues that affect two userstyles, though they are a rarity...

Any userstyle that involves a word with another userstyle would be at risk... I don't see why not.

isabelroses:
> Any userstyle that involves a word with another userstyle would be at risk... I don't see why not.

I know what your getting at but if we match by `^` and `$` we can no longer match `lbl:whatsapp-web, lbl:google`

uncenter:
> > Any userstyle that involves a word with another userstyle would be at risk... I don't see why not.
> 
> I know what your getting at but if we match by `^` and `$` we can no longer match `lbl:whatsapp-web, lbl:google`

Fair. Didn't even realize you could select multiple. Then we need a more complicated regex...

isabelroses (approved):

PR #982, opened 2024-06-13
build: use `npm:` instead of `esm.sh`

sgoudham:
Renovate is having some trouble parsing this as it's looking for it in the npm registry. Changing it to the `npm:` specifier should hopefully fix this.

When the deno-lib is brought into userstyles, we can look towards reverting this change to put more dependencies into the cache.


uncenter (approved):
1644-1648 should probably be removed? I know we said

PR #1008, opened 2024-06-22
fix(codeberg): more conditional changes to apply patches

isabelroses:
## 🗒 Checklist 🗒

### DIFF
--- a/styles/codeberg/catppuccin.user.css
+++ b/styles/codeberg/catppuccin.user.css
@@ -105,21 +105,17 @@
     }
   }
 
-  [data-theme="codeberg-light"] {
-    #catppuccin(@lightFlavor, @accentColor);
-  }
-  
-  [data-theme="codeberg-dark"] {
-    #catppuccin(@darkFlavor, @accentColor);
-  }
-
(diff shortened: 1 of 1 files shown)


styles/codeberg/catppuccin.user.css
```diff
@@ -105,14 +105,22 @@
...
+    [data-theme="codeberg-auto"] {
       #catppuccin(@lightFlavor, @accentColor);
     }
   }
 
   @media (prefers-color-scheme: dark) {
-    :root {
+    [data-theme="codeberg-auto"] {
       #catppuccin(@darkFlavor, @accentColor);
     }
```
uncenter, line 125: This won't work with our selectors above for Gitea in general? Above it just does prefers-color-scheme... so if you prefer dark but have Codeberg set to light you'll have a mismatch.

uncenter (changes requested):

styles/codeberg/catppuccin.user.css
isabelroses: thats fair

uncenter (approved):

Issue #781, opened 2024-04-08
nixos.wiki is now wiki.nixos.org


isabelroses:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:nixos.wiki

### Describe your problem.

https://nixos.wiki is being replaced by https://wiki.nixos.org, both of these websites use mediawiki but somewhat different css.

### What browser(s) are you seeing the problem on?

N/A

somerand0mcat:
i think #1120 will close this issue

orxngc:
Going to attempt to do this now.

PR #1050, opened 2024-07-06
fix(nixos-wiki): unthemed search bar

somerand0mcat:
i was doing some testing with the nixos.wiki userstyle and i found out that the search bar wasn't themed so i decided to fix it

image
image


styles/nixos.wiki/catppuccin.user.css
```diff
@@ -322,6 +322,40 @@
...
+      background-color: @base;
+      border: none;
+    }
+    
+    .search-types .current a {
+      color: @text;
+    }
+    
+    .oo-ui-inputWidget-input {
+       background-color: @subtext1 !important;
```
isabelroses, line 336: we shouldn't make subtext a background color

styles/nixos.wiki/catppuccin.user.css
```diff
@@ -322,6 +322,40 @@
...
+       color: @base !important;
+    }
+
+    & when (@lookup = latte) {
+     .oo-ui-inputWidget-input {
+       background-color: @base !important;
+       color: @text
+      }
+     .oo-ui-buttonElement-button {
+       background-color: @blue !important;
```
isabelroses, line 355: perhaps accent instead

styles/nixos.wiki/catppuccin.user.css
```diff
@@ -322,6 +322,40 @@
...
+       border: none;
```
isabelroses, line 356:
```suggestion
```
we don't want an opinionated change like this.

styles/nixos.wiki/catppuccin.user.css
```diff
@@ -322,6 +322,40 @@
     #footer {
       background-color: @mantle;
     }
+    
+    .mw-search-profile-tabs, {
+      background-color: @base;
+      border: none;
```
isabelroses, line 328:
```suggestion
```
we don't want an opinionated change like this.

styles/nixos.wiki/catppuccin.user.css
```diff
@@ -322,6 +322,40 @@
...
+    }
+    
+    .oo-ui-inputWidget-input {
+       background-color: @subtext1 !important;
+       color: @base;
+    }
+    
+    .oo-ui-buttonElement-button {
+       background-color: @blue !important;
+       border: none;
```
isabelroses, line 342:
```suggestion
```
we don't want an opinionated change like this.

styles/nixos.wiki/catppuccin.user.css
```diff
@@ -322,6 +322,40 @@
...
+     }
+    
+    .oo-ui-labelElement-label {
+       color: @base !important;
+    }
+
+    & when (@lookup = latte) {
```
isabelroses, line 349: try make latte the same as the dark flavours otherwise should be fine...

isabelroses (changes requested):

uncenter:
Seeing as nixos.wiki is no longer the suggested wiki site (see https://github.com/catppuccin/userstyles/issues/781) I'm not inclined to accept fixes for this userstyle. Thoughts @isabelroses?

isabelroses:
> Seeing as nixos.wiki is no longer the suggested wiki site (see #781) I'm not inclined to accept fixes for this userstyle. Thoughts @isabelroses?

I don't see too much harm in accepting fixes but ideally we should push for a new theme

styles/nixos.wiki/catppuccin.user.css
somerand0mcat: the search bar breaks when i switch to latte so

somerand0mcat: i'll keep it

isabelroses (changes requested):
1) Rember to update the version number

2) These update seems to break some stuff here
image
image

somerand0mcat:
dw i'll fix it

somerand0mcat:
@isabelroses fixed
image

uncenter:
See lint error.

somerand0mcat:
@uncenter fixed

somerand0mcat:
i also added some border colors to make it look pretty good
image

isabelroses (approved):

PR #1051, opened 2024-07-07
fix(codeberg): unthemed right action view

somerand0mcat:
the action view on the right was unthemed so i fixed it

image
image

### DIFF
--- a/styles/codeberg/catppuccin.user.css
+++ b/styles/codeberg/catppuccin.user.css
@@ -79,13 +79,18 @@
       }
     }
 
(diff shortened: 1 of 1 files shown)


isabelroses:
I believe these changes should be upstreamed to catppuccin/gitea. Since I can reproduce them on a diffrent forgejo server.

somerand0mcat:
> I believe these changes should be upstreamed to catppuccin/gitea. Since I can reproduce them on a diffrent forgejo server.

alright i'll theme it there ig

PR #1052, opened 2024-07-07
fix(arch-wiki): fix a bunch of broken stuff

somerand0mcat:
i was testing out the arch wiki and i found a bunch of broken stuff
image
image
which i fixed
image
image

### DIFF
--- a/styles/arch-wiki/catppuccin.user.css
+++ b/styles/arch-wiki/catppuccin.user.css
@@ -88,8 +88,10 @@
     .mw-parser-output {
       background-color: @base !important;
     }
+    
     .mw-page-container {
-      background-color: @mantle !important;
+      background-color: @base !important;
+      shadow: none;
     }
 
     .vector-header-container {
@@ -392,11 +394,6 @@
       color: @blue;
     }
 
-    .mw-body {
-      background-color: @base !important;
-      color: @text;
-    }
-
     .cdx-text-input__input:enabled {
       background-color: @surface0 !important;
       color: @text !important;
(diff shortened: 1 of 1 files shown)


styles/arch-wiki/catppuccin.user.css
```diff
@@ -429,6 +425,42 @@
...
+      background-color: @base !important;
+      color: @text !important;
+      border-color: @base !important;
+    }
+      
+    .mw-body {
+      background-color: @mantle !important;
+      color: @text !important;
+      border-color: @base !important;
+      border-radius: 10px;
```
isabelroses, line 439: Let's avoid opinionated changes like this
```suggestion
```

styles/arch-wiki/catppuccin.user.css
```diff
@@ -429,6 +425,42 @@
...
+      background-color: @mantle !important;
+    }
+      
+    .vector-sticky-pinned-container::after {
+      background: @base !important;
+    }
+      
+    .cdx-search-input {
+      border-color: @mantle !important;
+      border-width: 3px
```
isabelroses, line 458: same as prior comment
```suggestion
```

styles/arch-wiki/catppuccin.user.css
```diff
@@ -429,6 +425,40 @@
...
     }
+      
+    .cdx-button {
+      background-color: @base !important;
+      color: @text !important;
+      border-color: @base !important;
+    }
+      
+    .mw-body {
+      background-color: @mantle !important;
```
isabelroses, line 436: Content that is in the middle should be base following the style guide, perhaps make the header crust and the outer background mantle?
```suggestion
      background-color: @base !important;
```

somerand0mcat: i think thats a good idea yeah

isabelroses (commented):
This is going to need some changes
image

somerand0mcat:
@isabelroses fixed
image

isabelroses (approved):

PR #1058, opened 2024-07-08
feat(searxng): use only 2 instances instead of 20+ instances

somerand0mcat:
## 🔧 What does this fix? 🔧
as said in the title

## 🗒 Checklist 🗒


ryanccn:
One wonders if it would be better to just have only one or two default instances for simplicity

somerand0mcat:
> One wonders if it would be better to just have only one or two default instances for simplicity

what if the user that uses this style is using another instance

isabelroses:
> what if the user that uses this style is using another instance that is not these two

Just add it. We have a little tutorial on how to do it on the readme.

styles/searxng/catppuccin.user.css
```diff
@@ -17,120 +17,9 @@
...
-domain("sx.catgirl.cloud"),
-url-prefix("https://timdor.noip.me/searx/"),
-url-prefix("https://trydex.tk/searxng/"),
-domain("www.gruble.de"),
-domain("www.webrats.xyz"),
-domain("xcxc.ml"),
-domain("xo.wtf") {
+@-moz-document
+domain("https://search.im-in.space/"), 
+domain("https://search.projectsegfau.lt/") {
```
uncenter, line 22: Possibly stupid question but why these two specifically?

somerand0mcat: > Possibly stupid question but why these two specifically?

i just googled the instances list and picked these two

somerand0mcat: also they're pretty stable too

uncenter: These two weren't even on the original list though?

uncenter: Do you use SearXNG?

somerand0mcat: a bit, yeah

uncenter: Maybe just take the top three or something? Also leave a comment above the `@-moz-document` detailing where these are from would be great - `/* Domains picked from https://searx.space/. */

somerand0mcat: that sounds like a great idea

somerand0mcat: @uncenter i did all of your suggestions

uncenter (approved):

ryanccn (approved):

PR #1072, opened 2024-07-12
fix(duckduckgo): partially unthemed popup

somerand0mcat:
## 🔧 What does this fix? 🔧
as said in the title

image
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/duckduckgo/catppuccin.user.css
+++ b/styles/duckduckgo/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name DuckDuckGo Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/duckduckgo
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/duckduckgo
-@version 0.2.4
+@version 0.2.5
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/duckduckgo/catppuccin.user.css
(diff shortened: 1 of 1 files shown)


styles/duckduckgo/catppuccin.user.css
```diff
@@ -297,6 +297,12 @@
       background: linear-gradient(rgba(40, 40, 40, 0), @surface0);
     }
 
+    /* popup */
+    .atb-new .badge-link__checkbox.badge-link__checkbox_checked {
+        background: @mantle;
+        color: @text;
+    }
```
uncenter, line 304: Are you sure there aren't variables we can use here? I'm seeing other already themed variables with `atb` in them which makes me think we could use vars here.

somerand0mcat: i didnt check to see if there was any variables which we can use, maybe you can theme these ``atb`` variables and check to see if they actually theme?

uncenter: Yes... I'm asking you to do that. You should *always* be checking for variables.

somerand0mcat: i checked the variables and theres no way we can use them here so this codes good

somerand0mcat:
merged with #1077

PR #1075, opened 2024-07-12
refactor(wikipedia): rewrite

uncenter:
Rewrites Wikipedia! Their new dark mode support means they finally switched to properly using vars, so this works!

### DIFF
--- a/styles/wikipedia/catppuccin.user.less
+++ b/styles/wikipedia/catppuccin.user.less
@@ -228,6 +228,10 @@
 
     a {
       color: @accent;
+
+      &:focus {
+        outline-color: @accent;
+      }
     }
 
     .mw-parser-output {
--- a/styles/wikipedia/catppuccin.user.less
+++ b/styles/wikipedia/catppuccin.user.less
@@ -16,7 +16,9 @@
 @var checkbox highlight-redirect "Highlight redirect links" 0
 ==/UserStyle== */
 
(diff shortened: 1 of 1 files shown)


uncenter:
As mentioned in Discord; this rewrite is pretty usable for the most part, and most of the unthemed stuff you'll find is due to whatever element not properly using vars. I'd expect them to continue fixing issues like that as the rollout progresses and I think we'll wait a bit before merging.

somerand0mcat:
wait so the port was so unstable it had to be rewritten 2-3 times?

uncenter:
> why does this port always gets rewritten?

This is noise, and it doesn't? It hasn't been rewritten a single time before this?

somerand0mcat:
> This is noise, and it doesn't? It hasn't been rewritten a single time before this?
oh wait i thought it was rewritten twice my bad

GitMuslim:
this is not colored btw
image

uncenter:
As I said:

> most of the unthemed stuff you'll find is due to whatever element not properly using vars

I don't plan to theme stuff not using vars. I expect they will fix issues like that one in the future.

GitMuslim:
Oh I see!

uncenter:
Ended up theming the root and home pages that don't look like they are getting variables any time soon. Theme is usable, don't see any unthemed areas except for inline stuff. Should be good to go.

isabelroses (dismissed):

mustafakhalaf-git:
Glad this was done.

PR #1091, opened 2024-07-19
refactor(scripts): use deno's std on jsr

uncenter:
https://deno.com/blog/std-on-jsr
Don't need #990 anymore.


isabelroses (dismissed):
checks pass, diff passes the eye test.

PR #1069, opened 2024-07-11
fix(pinterest): unthemed variables

somerand0mcat:
## 🔧 What does this fix? 🔧
as said in the title

## 🗒 Checklist 🗒

- [X] I have read and followed Catppuccin's [contributing guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md).
- [X] I have updated the version appropriately in the `==UserStyle==` header of the `catppuccin.user.css` file.

PR #1101, opened 2024-07-21
fix(pinterest): unthemed variables

somerand0mcat:
## 🔧 What does this fix? 🔧
exact same thing as #1069
i couldn't reopen that pr because the branches were deleted

## 🗒 Checklist 🗒


isabelroses (changes requested):
can we get a version bump, otherwise looks good

isabelroses (approved):

PR #1103, opened 2024-07-21
fix(modrinth): cleanup

uncenter:


thismoon:
@uncenter I think this broke #1099 ? I can't verify what happened exactly (I'm on phone rn so no devtools) but the logo that was colored before isn't colored anymore



(Note: this is the logo in the bottom of the page where there's a button for the blog, not the top banner)

uncenter:
Huh weird it was working exactly the same before on my end, and now I'm seeing that as well. I quadruple checked!

uncenter:
The original CSS isn't working either!

PR #1115, opened 2024-07-24
fix(pypi): unthemed text

somerand0mcat:
## 🔧 What does this fix? 🔧
as said in the title
image
image

## 🗒 Checklist 🗒


styles/pypi/catppuccin.user.css
```diff
@@ -135,6 +135,10 @@
       background-image: linear-gradient(90deg, @surface0, @surface0) !important;
     }
 
+    .project-description blockquote {
+      color: @text !important;
```
trinkey, line 139: i think this should be subtext0 as in the default styles its a more muted color than the main text

trinkey (changes requested):

trinkey (approved):

PR #1183, opened 2024-08-10
fix(pronouns.page): disabled buttons

somerand0mcat:
## 🔧 What does this fix? 🔧
as said in the title
image
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/pronouns.page/catppuccin.user.css
+++ b/styles/pronouns.page/catppuccin.user.css
@@ -212,7 +212,7 @@
       background-color: @accent-color !important;
     }
 
-    button:disabled {
+    .btn:disabled {
        background-color: darken(@accent-color, 5%);
        color: @crust;
     }
--- a/styles/pronouns.page/catppuccin.user.css
+++ b/styles/pronouns.page/catppuccin.user.css
@@ -104,7 +104,7 @@
       --bs-btn-active-bg: @accent-color;
       --bs-btn-active-border-color: @accent-color;
       --bs-btn-disabled-color: @accent-color;
-      --bs-btn-disabled-bg: transparent;
+      --bs-btn-disabled-bg: darken(@accent-color, 5%);
       --bs-btn-disabled-border-color: @accent-color;
     }
 
@@ -210,11 +210,6 @@
 
     .btn-square {
       background-color: @accent-color !important;
-    }
-
-    .btn:disabled {
-       background-color: darken(@accent-color, 5%);
-       color: @crust;
     }
 
     /* Inputs */
--- a/styles/pronouns.page/catppuccin.user.css
+++ b/styles/pronouns.page/catppuccin.user.css
@@ -212,8 +212,8 @@
       background-color: @accent-color !important;
     }
 
-    button[disabled*="disabled"] {
-       background-color: darken(@accent-color, 5);
+    button:disabled {
+       background-color: darken(@accent-color, 5%);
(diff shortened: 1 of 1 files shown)


styles/pronouns.page/catppuccin.user.css
```diff
@@ -212,6 +212,11 @@
       background-color: @accent-color !important;
     }
 
+    button[disabled*="disabled"] {
+       background-color: darken(@accent-color, 5);
+       color: @crust;
+    }
```
uncenter, line 218: Does `button:disabled` work the same here?

uncenter: Also please use the percent symbol with Less functions (`5%` not `5`).

somerand0mcat: i forgot about ``:disabled`` yeah

styles/pronouns.page/catppuccin.user.css
```diff
@@ -212,6 +212,11 @@
       background-color: @accent-color !important;
     }
 
+    button:disabled {
+       background-color: darken(@accent-color, 5%);
+       color: @crust;
+    }
```
uncenter, line 218: This actually looks wrong. Look at how buttons and disabled buttons/colors are themed generally in the lines above.

somerand0mcat: what if we use ``.btn`` or ``.btn-primary`` instead of ``button``?

uncenter: Right can you move and adjust this there? The selector they use to theme disabled buttons should be the same we use in most cases.

somerand0mcat: @uncenter uh sure

somerand0mcat: @uncenter wait what if we can just use variables instead of manually theming the disabled buttons?

uncenter: "uh sure ig" after me telling you yes, you should "use .btn or .btn-primary instead of button for this" (your words), is wild.

somerand0mcat: > "uh sure ig" after me telling you yes, you should "use .btn or .btn-primary instead of button for this" (your words), is wild lmao.

its because i didn't understand some parts of your comment lol

somerand0mcat: also i dug deeper into the code of this userstyle and found out you only need ``--bs-btn-disabled-bg`` to theme the disabled buttons, can't we just use that instead of manually theming the disabled buttons?

uncenter: Yep that's what I meant by adjusting. In the future you should *always* check for things like that - don't just theme the first thing that "fixes" it, go back up the element tree and find the one with the changes/miscoloration and see what caused it.

somerand0mcat: yeah i'll do that in the near future, im also going to follow some feedback from bell too

uncenter: I hope so!

somerand0mcat: > I hope so!

im also gonna try not reverting back after following the feedback, which is probably why i got banned from the discord

uncenter (approved):

PR #1184, opened 2024-08-11
fix(google-photos): text, search bar and icons

somerand0mcat:
## 🔧 What does this fix? 🔧
as said in the title
i would put a before/after in here but it would take way too much space because of how many unthemed stuff there is, so i decided not to 

## 🗒 Checklist 🗒


GenShibe:
can you give me a before & after so i have a better idea of what's changed ty

somerand0mcat:
> can you give me a before & after so i have a better idea of what's changed ty
the readme should have a before/after now

GenShibe (approved):
LGTM

Issue #1187, opened 2024-08-12
error found in the code


pdimu:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:mastodon

### Describe your problem.

when opening the theme in stylus, it points out an error at [line 29](https://github.com/catppuccin/userstyles/blob/main/styles/mastodon/catppuccin.user.css) and the theme doesn't load on mastodon

### Attach screenshots.

![image](https://github.com/user-attachments/assets/a8c56cbe-6018-41db-9e4e-27b8379981e6)

### What browser(s) are you seeing the problem on?

G-Chrome 127.0.6533.100 (Official Build) (64-bit) (cohort: Stable) 

### Any additional comments?

i don't know to fix tbh

uncenter:
The latest version of the theme loads fine for me. The "unknown word" issue does not block the loading of the userstyle and isn't fixable on our end unless we want to rename the theme...

PR #1189, opened 2024-08-12
fix(scripts/import-styles): disable stylelint linter in stylus editor

uncenter:
Helps manage #1187.


pdimu:
@uncenter what does this actually changing and how it is a fix?

uncenter:
As said in the title, it disables the stylelint linter in Stylus's editor when importing from our generated all userstyles import file. This gets rid of that error.

isabelroses (approved):

PR #1207, opened 2024-08-16
feat(namemc): init

Nyatalieeee:
## 🎉 Theme for [NameMC](https://namemc.com/) 🎉

NameMC is a website for viewing Minecraft skins, servers, usernames, players, etc.

skins
names
servers
capes
search
profile

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [x] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [x] I have ensured that the new directory is in **lower-kebab-case**.
  - [x] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have made sure to update the
      [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml)
      file with information about the new userstyle.
- [x] I have included the following files:
  - [x] `catppuccin.user.css` - all the CSS for the userstyle, based on the
        template.
  - [x] `preview.webp` - composite image of all four individual flavor screenshots (taken with the default accent color of mauve) stitched together, generated via [Catwalk](https://github.com/catppuccin/catwalk).

PR #1249, opened 2024-08-31
fix(namemc): requested changes

Nyatalieeee:
## 🔧 What does this fix? 🔧

fixes the issues highlighted after #1207 was mistakenly merged

## 🗒 Checklist 🗒

### DIFF
--- a/styles/namemc/catppuccin.user.css
+++ b/styles/namemc/catppuccin.user.css
@@ -285,15 +285,10 @@
     .checkered {
       @dark: fade(@surface0, 40%);
       @light: fade(@surface1, 40%);
-      @svg: escape('<svg width="64" height="64" xmlns="http://www.w3.org/2000/svg"><path fill="@{dark}" d="M0 0h32v32H0z"/><path fill="@{light}" d="M32 0h32v32H32zM0 32h32v32H0z"/><path fill="@{dark}" d="M32 32h32v32H32z"/></svg>');
+      @svg: escape(
+        '<svg width="64" height="64" xmlns="http://www.w3.org/2000/svg"><path fill="@{dark}" d="M0 0h32v32H0z"/><path fill="@{light}" d="M32 0h32v32H32zM0 32h32v32H0z"/><path fill="@{dark}" d="M32 32h32v32H32z"/></svg>'
+      );
       background-image: url("data:image/svg+xml,@{svg}");
-    }
-
-    /* main page server list links */
-    table.server-table {
-      > tbody > tr > td > a {
-        color: @accent-color !important;
-      }
     }
 
     /* skins page nav buttons */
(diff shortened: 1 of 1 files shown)


styles/namemc/catppuccin.user.css
```diff
@@ -52,12 +52,59 @@
...
+      --bs-btn-active-color: @hover;
+    }
+
+    #rgbify(@color) {
+      @rgb-raw: red(@color), green(@color), blue(@color);
+    }
+
     --bs-body-color: @text;
-    --bs-body-color-rgb: @text;
+    --bs-body-color-rgb: #rgbify(@text);
```
uncenter, line 103: You need to unfortunately add these brackets after every #rgbify() usage:
```suggestion
    --bs-body-color-rgb: #rgbify(@text) [];
```

styles/namemc/catppuccin.user.css
```diff
@@ -277,6 +277,7 @@
       background-color: @mantle !important;
     }
 
+    /* main page server list links */
```
uncenter, line 292: I don't think this section is needed in the first place then?

CleanShot 2024-09-07 at 13 37 35

It would already be themed by the themed variable shown in that screenshot.

styles/namemc/catppuccin.user.css
```diff
@@ -358,17 +358,12 @@
...
+        td a {
           color: @accent-color !important;
         }
-
-        &.border-top {
-          td a {
-            color: @accent-color !important;
-          }
-        }
       }
```
uncenter: I don't think any of this section is needed. Try removing it on your end, I'm not seeing any visible changes. Looks like maybe that link is already themed?

styles/namemc/catppuccin.user.css
Nyatalieeee: AH good catch, i'll get rid of that section shortly

uncenter (changes requested):

uncenter (changes requested):
Can we change `--bs-success-border-subtle` and `--bs-success-text-emphasis` to `@green`? And then `--bs-success-bg-subtle` to `fade(@green, 50%)` or something? This better themes the "Requires emerald rank" boxes on https://namemc.com/my-profile/info (I'm not sure if 50% is the right opacity to fade for the bg, please check and change if necessary ofc).

uncenter:
Along with the success boxes, can we change `--bs-warning-border-subtle` to `@yellow` (to match the text color in the center) and `--bs-warning-bg-subtle` to `fade(@yellow, 50%)` (or with whatever opacity you deem fit, should be the same as the one for success above). This themes this box on the login page better:
CleanShot 2024-09-07 at 13 50 37

uncenter:
Sorry, on the last two: just tried 50% and it is too much. 20% looks good to me.

styles/namemc/catppuccin.user.css
Nyatalieeee: yeah it looks like what's happening here is _likely_ the same as the previous one; overriding the variable that does the exact same thing :P

uncenter (approved):
Thank you so much! Looks much better now, I have nothing else. Great work on the userstyle!

PR #1269, opened 2024-09-06
docs(CONTRIBUTING.md): link directly to the Stylus wiki

thismoon:
in https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md there's a hyperlink that links to https://github.com/catppuccin/userstyles/blob/main/docs/tips-and-tricks.md which links to https://github.com/openstyles/stylus/wiki/Writing-UserCSS without providing any additional information. therefore i think it's better to save a click and link directly to the stylus wiki. that is unless this was done purposefully so users know the existence of the tips and tricks file?


uncenter:
A bit of what you said, purposeful to bring awareness to the file, but also if we ever have the need to include additional information (or the original if that wiki were to ever change) we can just edit that tip.

uncenter:
I will be restructuring the docs situation soon (https://github.com/catppuccin/catppuccin/issues/2506) though, so I might change it then (tips would be in separate files I think, so it would make less sense).

thismoon:
so do i close this pr?

uncenter:
Thanks for bringing it up though! I'll keep it in mind as previously mentioned when making those docs changes.

PR #1270, opened 2024-09-06
fix(youtube): misc elements and adjust background colors

thismoon:
## 🔧 What does this fix? 🔧

- changed `--yt-spec-raised-background` from `@base` to `@surface0` (because the search suggestions background should be a bit brighter than the main background)
- theme the video duration on the thumbnails
- make the description background darker (same color as the buttons like in unthemed youtube) by changing `--yt-spec-badge-chip-background` to `@surface0` on non-latte flavors
- make the enabled menu item switch background colored with the accent color

## 🗒 Checklist 🗒

### DIFF
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name YouTube Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/youtube
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/youtube
-@version 4.2.4
+@version 4.2.5
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/youtube/catppuccin.user.css
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Ayoutube
 @description Soothing pastel theme for YouTube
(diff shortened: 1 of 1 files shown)


uncenter:
I hope you don't mind but I applied some more changes that were too annoying to fit in a comment diff. I changed the value of `--yt-spec-static-overlay-background-heavy`, which themes the other thumbnail icons like add to watch later and add to queue (and hopefully nothing else...). I also increased the opacity of the timestamp background a little bit.

isabelroses:
can i see some before and afters so i know what I'm looking at. thanks.

thismoon:
@isabelroses
# search bar
before:
image
after:
image
# thumbnail overlay
before:
image
after:
image
# description background
before:
image
after:
image
# item switch background
before:
image
after:
image

uncenter (approved):

isabelroses (approved):

uncenter:
Changing `--yt-spec-raised-background` actually messes up the shorts comments section - going to have to revert that bit.
image

thismoon:
@uncenter by messes up i assume that you mean that the background is very bright? because i don't see anything that became invisible/hard to read
if you check the unthemed youtube, the background of the comments popup is actually brighter than the normal background (because it uses the raised variable)
image
and obviously the raised variable should be brighter/different that the base one
also reverting the change messes up that sidebar next to the short (the one containing the like and comment buttons), the search suggestions list, the arrow buttons on any vertically scrollable sections like in a channel's homepage...
you should've made an exception to theme the shorts comments with `@base` or at least changed the variable to something like `lighten(@base, 25%)` and darken for latte etc

uncenter:
I find that the base themed search dropdown and comments sidebar look better.

thismoon:
i honestly don't like how the search suggestions bar have the same background as the main page. it makes it not have a sense of depth although it's another element on top of the main page. it needs to have a different background to look on top/separate (some other websites would add an outline around the raised element but we aren't going that far)

thismoon:
@uncenter i am still waiting for an answer on if we should ask the community for an opinion using a poll.

uncenter:
I don't think we would get many responses. Go ahead and raise a PR I guess and we can discuss there. We'll have to do a more thorough review to make sure it doesn't break other areas.

PR #1271, opened 2024-09-07
chore: add .envrc for direnv

HeitorAugustoLN:
This adds an `.envrc` file to automatically use the nix flake dev shell when entering this repository (for direnv users)

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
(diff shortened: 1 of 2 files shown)


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
isabelroses, line 1: lets make it so that it only works with nix direnv installs
```suggestion
if has nix_direnv_version; then
  use flake
fi
```

isabelroses (approved):

PR #1283, opened 2024-09-10
fix(crates.io): misc

somerand0mcatt:
## 🔧 What does this fix? 🔧
Fixes a unthemed button aswell as links in the wrong color.

<details>
<summary>Before & After (Outdated)</summary>

</details>

## 🗒 Checklist 🗒

### DIFF
--- a/styles/crates.io/catppuccin.user.css
+++ b/styles/crates.io/catppuccin.user.css
@@ -265,14 +265,12 @@
       }
     }
 
-    /* Report button */
-    [class*="_report-button_"] {
+    [class*="_red-button_"] {
       color: @crust;
-      background: @red;
-
-      &:hover {
-        background-color: darken(@red, 5%);
-      }
+      --bg-color-top: @red;
+      --bg-color-bottom: @red;
+      --bg-color-top-dark: darken(@red, 5%);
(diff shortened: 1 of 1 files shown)


uncenter:
In summary: figure out which line and why it is breaking things, don't just remove it! What was it trying to theme? Make sure that is then still themed, and then go and update the class selector for the *red* button report button. You can also just set the variables on that class selector instead of setting `color:` and whatnot.

somerand0mcatt:
> You've missed noticing the key bug here. The reason the text color of the button is text/white is because something changed and a line I had in here broke things.

I did a bit of searching and found out some code for the header buttons/text is actually the reason why the colors of the button are just text/white, I'll fix it later when I have time. (Same with the report button issue)

> And if you look further, I don't know why you are theming just the report button. The original CSS on the website is theming red buttons with the class [class*="_red-button_"] and using variables within that.

I didn't notice nether the red button class nor the red button CSS so thats the reason why I only themed the report button, For the  variables I'll fix it when I have some time to.

Also @uncenter, can you make it so I can get notifications and react to your messages? I think PRs might get a bit more harder without these.

uncenter (approved):

PR #1286, opened 2024-09-11
fix(hacker-news): set body text color

uncenter:
## 🔧 What does this fix? 🔧

Themes some text on the login page which is just the page default, setting it here works and presumably fixes other text like it.

## 🗒 Checklist 🗒

- [ ] I have updated the version appropriately in the `==UserStyle==` header of the `catppuccin.user.css` file.


lucasmelin (approved):
LGTM, thanks! ⭐

PR #1291, opened 2024-09-11
fix(modrinth): unthemed elements

thismoon:
## 🔧 What does this fix? 🔧

theme unthemed elements (especially in the landing page)

## 🗒 Checklist 🗒

### DIFF
--- a/styles/modrinth/catppuccin.user.css
+++ b/styles/modrinth/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name Modrinth Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/modrinth
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/modrinth
-@version 1.2.8
+@version 1.2.9
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/modrinth/catppuccin.user.css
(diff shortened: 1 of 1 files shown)


isabelroses (approved):

uncenter:
The text of this sign up button and the buttons/links in the header on hover is unthemed:

CleanShot 2024-09-19 at 21 40 01
CleanShot 2024-09-19 at 21 40 28

Would you mind fixing this as well?

isabelroses (approved):

styles/modrinth/catppuccin.user.css
```diff
@@ -165,6 +176,17 @@
       border-color: @text transparent transparent;
       color: @text;
     }
+    // header button hover
+    .btn-wrapper:hover {
+      a,
+      button {
+        --tw-brightness: brightness(1) !important;
```
uncenter, line 183: I think I would prefer `filter: none;` to be more explicit here.

styles/modrinth/catppuccin.user.css
```diff
@@ -165,6 +176,17 @@
...
+      }
+    }
+    // loading indicator
```
uncenter, line 186:
```suggestion
```
Generally, no need for comments like these when the name of the class/id/attr/etc is self explanatory (in this case the class contains both words of the comment).

uncenter (approved):

PR #1292, opened 2024-09-11
fix(stylus): delete button

thismoon:
### fix the delete button in the popup
before
image
after
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/stylus/catppuccin.user.css
+++ b/styles/stylus/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name Stylus Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/stylus
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/stylus
-@version 1.2.5
+@version 1.2.6
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/stylus/catppuccin.user.css
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Astylus
 @description Soothing pastel theme for Stylus
--- a/styles/stylus/catppuccin.user.css
+++ b/styles/stylus/catppuccin.user.css
@@ -276,7 +276,7 @@
       }
 
       > div,
-      & [data-cmd="delete"] {
+      [data-cmd="delete"] {
         border-color: @red;
       }
     }
--- a/styles/stylus/catppuccin.user.css
+++ b/styles/stylus/catppuccin.user.css
@@ -276,7 +276,7 @@
       }
 
       > div,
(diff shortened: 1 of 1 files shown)


styles/stylus/catppuccin.user.css
```diff
@@ -276,7 +276,7 @@
       }
 
       > div,
-      &[data-cmd="delete"] {
+      & [data-cmd="delete"] {
```
uncenter, line 279:
```suggestion
      [data-cmd="delete"] {
```

uncenter:
CleanShot 2024-09-11 at 14 28 44

I don't see that yellowish bar from the first image?

thismoon:
@uncenter can you inspect what class that bar is using (so we can see if it's already in the theme)? i think we're looking at different bars
the one i fixed should appear when editing a catppuccin userstyle using the addon and has a `applies-to` class
this is the rule it had before my fix
```css
.applies-to {
	background-color: rgba(44, 44, 44, 1);
	border-top: 1px solid rgba(153, 153, 153, 0.2);
	border-bottom: 1px solid rgba(153, 153, 153, 0.2);
}
```

uncenter:
CleanShot 2024-09-11 at 15 39 35

thismoon:
im confused
can you find where is the background color coming from?
the theme only has
```css
    .applies-to input,
    .applies-to select {
      background: @base;
      border-color: @surface0;
      color: @subtext0;
    }
```
it's using the rgb mantle color from the macchiato flavor. but usually userstyle colors (for me at least) are hex codes
maybe you have my version enabled accidentally?

uncenter:
I definitely don't have your version installed yet it is themed. I'm lost too. @isabelroses can you repro the unthemed bar here?

isabelroses:
> I definitely don't have your version installed yet it is themed. I'm lost too. @isabelroses can you repro the unthemed bar here?

I can't seem to repro either...

thismoon:
weird
i still didn't get confirmation on where is the `.applies-to` background color coming from

somerand0mcatt:
I think i can eeproduce that, I've seen that ``.applies-to`` bar before.

somerand0mcatt:
I can't reproduce it myself too, I've tried different browsers, tested different versions of the userstyle (1.2.0+), it still didn't work.
I think somethings probably overriding that background (or its just a browser issue).

thismoon:
i found out the reason
the unthemed bar only appear in private browsing (firefox at least)
image
would this be a valid reason to keep it (and theme the text and borders too because they're unthemed) or do i consider it a stylus bug and undo the fix?

uncenter (approved):

PR #1303, opened 2024-09-13
fix(duckduckgo): light theme flash on page load

thismoon:
## 🔧 What does this fix? 🔧

on duckduckgo when the page loads and before js finished loading the page have a `no-theme` class.
on unthemed duckduckgo this class makes the page follows the system theme
but with catppuccin that defaults to the light mode, which causes a light flash for a short time on load

## 🗒 Checklist 🗒


GenShibe (approved):
lgtm, ship it 👍🏼

PR #1367, opened 2024-10-09
fix(youtube): live ring

isabelroses:
## 🔧 What does this fix? 🔧

fixes new youtube live ring
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/youtube/catppuccin.user.css
+++ b/styles/youtube/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name YouTube Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/youtube
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/youtube
-@version 4.2.12
+@version 4.2.13
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/youtube/catppuccin.user.css
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Ayoutube
 @description Soothing pastel theme for YouTube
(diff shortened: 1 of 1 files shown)


uncenter (changes requested):
I can't find where this appears, but I did find two other cases where the ring is present that are unthemed:

1. On the profile/channel page of an account that is currently live:

image

2. A similar, smaller circle ring (hopefully/presumably the same component or class) under the video tile (this one was from the subscriptions page, but it would be the same if someone was live appeared on your suggestions/home page.

image

isabelroses:
> I can't find where this appears, but I did find two other cases where the ring is present that are unthemed:
> 
> 1. On the profile/channel page of an account that is currently live:
> 
> image
> 
> 2. A similar, smaller circle ring (hopefully/presumably the same component or class) under the video tile (this one was from the subscriptions page, but it would be the same if someone was live appeared on your suggestions/home page.
> 
> image

First channel i found and it was working. Perhaps weird region difference?

{2ED8EC23-27F0-464B-883B-01CD12C3B380}
{6E22B74D-8907-4C63-943F-0812EC3456A0}

styles/youtube/catppuccin.user.css
```diff
@@ -472,6 +472,17 @@
...
+    /* youtube live ring */
+    .yt-spec-avatar-shape__badge-text {
+      color: @crust;
+    }
+    .yt-spec-avatar-shape--live-ring {
+      border-color: @accent-color;
+    }
+    .yt-spec-avatar-shape__live-badge {
+      background-color: @accent-color;
+    }
```
uncenter, line 484: Can we put this somewhere more relevant, like near the other live/premiere badges/stuff is?

uncenter (approved):

PR #1372, opened 2024-10-15
fix(wikipedia): missing infobox classes

arminius-smh:
## 🔧 What does this fix? 🔧

Hello,
this pr fixes unthemed infobox subheaders, as the heading's text wasn't readable before

Here is an example site that has these classes: https://en.wikipedia.org/wiki/Flying_Witch

before:
wikipedia_before

now:
wikipedia_after

## 🗒 Checklist 🗒


isabelroses (approved):
Thanks for the PR

PR #1373, opened 2024-10-15
fix(mastodon)!: update applied domains

GenShibe:
## 🔧 What does this fix? 🔧
refactors the default domains to the three provided by hammy

## 🗒 Checklist 🗒


sgoudham (commented):
Mastodon social makes sense to have, fosstodon makes sense to have (for now?) as we have a prior account on there and social.catppuccin.com as that's now our official instance.

I didn't really want to set the precedent that we should add "random" domains into the base userstyle.

isabelroses (approved):

sgoudham (approved):

PR #1418, opened 2024-11-03
fix(hackage): color the synopsis toggle and older Haddock pages

jn-sena:
## 🔧 What does this fix? 🔧

* Colored the MathJax renders using proper filters instead of fully inverting the color.
* Fixed the synopsis toggle in Haddock using filters too.

the synopsis toggle before the change
the synopsis toggle after the change

* Themed older Haddock pages ([example](https://hackage.haskell.org/package/base-4.0.0.0/docs/Data-List.html)) and sources.

old Haddock page before the change
old Haddock page after the change

## 🗒 Checklist 🗒

### DIFF
--- a/styles/hackage/catppuccin.user.css
+++ b/styles/hackage/catppuccin.user.css
@@ -103,7 +103,13 @@
     h4,
     h5,
     h6,
-    summary {
+    summary,
+    .section1,
+    .section2,
+    .section3,
+    .section4,
+    .section5,
+    .section6 {
       color: overlay(@accent-color, @surface0);
       filter: none;
     }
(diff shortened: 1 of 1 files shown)


isabelroses (approved):

styles/hackage/catppuccin.user.css
```diff
@@ -242,8 +285,9 @@
     }
 
     #synopsis {
-      summary {
-        color: transparent;
+      summary,
+      #control\.syn {
+        filter: @synopsis-filter;
```
uncenter, line 290: What color is this `@synopsis-filter` supposed to be?

jn-sena: `surface0` as the background, though it also changes the text similar to just inverting it so it's not ideal. I think that was why I didn't name it `@surface0-filter`.

uncenter: So is it not truly a surface0 filter then? Can you link me the page you can find this element on and maybe we can use an SVG approach or similar? I would like to move away from any nonstandard filters in #1547 now that we have https://github.com/catppuccin/userstyles/blob/main/docs/guide/images-and-svgs.md#non-svg-images-or-many-img-elements-with-external-svgs.

Issue #1415, opened 2024-11-01
Pinterest: unthemed icons and searchbar


duckaleuze:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:pinterest

### Describe your problem.

The theme is broken

### Attach screenshots.

![image](https://github.com/user-attachments/assets/8aa56209-7c77-47a2-aa5e-af19b9775cd9)

### What browser(s) are you seeing the problem on?

Firefox 132.0

nuexq:
the nav bar icons, search bar and search guides has been fixed by #1581

PR #1405, opened 2024-10-27
fix(pinterest): unthemed elements

thismoon:
## 🔧 What does this fix? 🔧

some unthemed elements and broken stuff

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [contributing guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md).
- [x] I have updated the version appropriately in the `==UserStyle==` header of the `catppuccin.user.css` file.

Issue #1301, opened 2024-09-13
Wrong colors in the Pinterest. Dark Macchiato, Teal.


IvanKarpov-1:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:pinterest

### Describe your problem.

The like icon when not liked is black so it's hard to see (first image).
The text on the "Save" button is light so it's hard to read (first image).
When hovering over the "Save" button, it turns red (second image). But, maybe it should turn, I don't know.
When the post is saved, the text in the pop-up message, where it is saved, is dark, so it's hard to read (third image).
When unsave the post, the text in the pop-up message is dark, so it's hard to read (fourth image).

### Attach screenshots.

First image:
![image1](https://github.com/user-attachments/assets/3a1dce5c-7df0-4b0e-a6ab-f594d48575b5)
Second image:
![image2](https://github.com/user-attachments/assets/8861c395-bb7c-4dca-8fb5-003a7e9a0aad)
Third image:
![image](https://github.com/user-attachments/assets/27ffebc0-e9b3-4e38-9ad2-e5f531f47468)
Fourth image:
![image](https://github.com/user-attachments/assets/e2ec8cc9-d7a5-4f13-9c09-c0e788cb4460)

### What browser(s) are you seeing the problem on?

Google Chrome 128.0.6613.119

### Any additional comments?

I'm using Catppuccin Macchiato Dark with a Teal accent color.

NeonGamerBot-QK:
i think these features may have existed at some point but since pintrest has annoying css the selectors broke. 
ill try my best to implement the fixes if possible.

nuexq:
some problems got fixed by the prev pull request, and other issues like the svg's of reaction button and others got fixed by #1581.

PR #1420, opened 2024-11-04
fix(pinterest): improve readability and fix unthemed elements

### LINKED ISSUE #1378: Wrong colors in the Pinterest. Latte, Pink.

- [X] I am using the latest version of the provided userstyle.

lbl:pinterest

### Describe your problem.

 in the first image he gets this horrible error.

### Attach screenshots.

Captura de tela 2024-10-16 130132

Google Chrome 129.0.6668.103

### Any additional comments?

I'm using Catppuccin Latte with a Pink accent color.

K1aymore:
## 🔧 What does this fix? 🔧

I copied the changes from #1405 into here just in case, I wasn't aware of it when I started working.

- Fixes #1378
- Fixes 4/5 issues in #1301
  - The text on the "Save" button is more readable in different accent colors
  - When hovering over the "Save" button, it dims rather than becoming red
  - When a post is saved, the text in the pop-up message is readable
  - When a post is unsaved, the text in the pop-up message is readable
- The icons when hovering over a pin (among others) now have a dark background in dark themes
- The dropdown of which board to save to is now more readable in dark themes
- The "Created" and "Saved" buttons on the main profile page are themed
- A strange box behind section previews is removed
- The "Invite Collaborators" buttons no longer have a white border around them
- Pin editing mode is themed nicer
- Hover tooltips are no longer inverted (possibly my preference)
- Search dropdown is now the correct color
- Pin view:
  - Fixed the shadow around the pin
  - Increased shadow strength
  - Rounded corners
  - "View larger" and "Image search" buttons properly themed
  - Field for comment entry is no longer transparent

I could not reproduce #1415 but it seems like that is fixed?

What still needs to be fixed:
- Creating new pins menu
- Profile editing menus
- Pin outlines when selected in organization mode and using a dark theme are hard to see

## 🗒 Checklist 🗒


K1aymore:
Some before and afters:

Before (random unsaved pin):
image
After:
image

Before (main profile page):
image
After:
image

Before (editing a pin):
image
After:
image

Before (the issue from #1378):
image
After:
image

styles/pinterest/catppuccin.user.css
```diff
@@ -193,12 +195,24 @@
...
+    // Profile button
+    // Invite collaborators buttons
+    div[style="border: 1px solid white;"],
     div[data-test-id="gestalt-avatar-svg"] {
-      box-shadow: 0 0 0 1px @crust !important;
+      box-shadow: 0 0 0 0 @surface1 !important;
+      outline-color: @base !important;
+      outline-width: 0 !important;
+      border-color: @base !important;
+      border-width: 1px !important;
```
isabelroses, line 206: do we need this line?

styles/pinterest/catppuccin.user.css
```diff
@@ -193,12 +195,24 @@
...
     }
 
+    // Profile button
+    // Invite collaborators buttons
+    div[style="border: 1px solid white;"],
     div[data-test-id="gestalt-avatar-svg"] {
-      box-shadow: 0 0 0 1px @crust !important;
+      box-shadow: 0 0 0 0 @surface1 !important;
+      outline-color: @base !important;
+      outline-width: 0 !important;
```
isabelroses, line 204: do we need this line?

isabelroses (changes requested):
thanks for the PR and fixing all those issues :)

styles/pinterest/catppuccin.user.css
K1aymore: I don't see any change from box-shadow, we can remove that.

styles/pinterest/catppuccin.user.css
K1aymore: border-color should be base, but we don't need to set border-width (and it should not be zero). We could either set outline-color to base or outline-width to zero, but we don't need both. Setting outline-color to base seems more consistent since it sets border-color as well.

K1aymore: It also doesn't look like we need to set background-color right here either, and setting text to subtext1 isn't consistent with the color of other icons I believe

isabelroses:
Thanks for the fix :)

PR #1433, opened 2024-11-09
fix(duckduckgo): unthemed elements

thismoon:
## 🔧 What does this fix? 🔧

- new accent color variable
image
- tooltip background color
image
- menu item hover background
image
- code background
[example page](https://duckduckgo.com/?q=python+how+to+split+a+string)
search page:
before:
image
after:
image

ai chat:
before:
image
after:
image

- this border
image
- unexpanded lyrics gradient and explicit tag
[example page](https://duckduckgo.com/?q=pride+kendrick+lyrics)
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/duckduckgo/catppuccin.user.css
+++ b/styles/duckduckgo/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name DuckDuckGo Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/duckduckgo
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/duckduckgo
-@version 0.3.0
+@version 0.3.1
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/duckduckgo/catppuccin.user.css
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Aduckduckgo
 @description Soothing pastel theme for DuckDuckGo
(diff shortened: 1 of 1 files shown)


styles/duckduckgo/catppuccin.user.css
```diff
@@ -448,6 +451,14 @@
     .module__inner-toggle::after {
       background-color: @surface2 !important;
     }
+    .module--lyrics:not(.is-expanded)
+      .module--lyrics__footer.can-expand::after {
+      background: linear-gradient(rgba(0, 0, 0, 0), @surface0);
```
uncenter, line 456: Can we avoid using a non-palette color in this gradient?

thismoon: i used it after checking that there are other non palette colors in the style
image
there are 6 other occurrences of using an rgba for an invisible color
it's an invisible color with 0 opacity so there's no difference between it and `fade(@surface0, 0%)`. also when reading the code `rgba(0, 0, 0, 0)` is clearer and feels more like an invisible color while the other form feels more of a palette color that has an opacity
but that's my opinion, if you really think that using fade with a palette color and zero percentage is more appropriate i'll change it

uncenter: Ah I didn't see it was 0. Why not just use `transparent` for clarity?

thismoon: oh i forgot that i can do that. i replaced all the transparent colors with `transparent`. is it alright now?

GenShibe (approved):

PR #1445, opened 2024-11-13
docs!: restructure documentation

### LINKED ISSUE #1089: Document the CSS color filter generator technique for theming single color images

### Describe your issue.

We [use this technique in various places](https://github.com/search?q=repo%3Acatppuccin%2Fuserstyles%20%22filter%3A%20%40%22&type=code) for simple. single color PNGs/non vector images. Quite useful for small things, but the color filters need to be generated every time and we don't even document this at the moment. I'm thinking we create a new filter palette that is like our hex palette but with filters. And this wouldn't be included in the template by default but could be added for the userstyles that need it.

uncenter:
Closes #1089. Thanks @Papweer for the generated palette! Much appreciated.


uncenter:
I think we still need a better way of surfacing the tips somehow, not sure what the best way is. When new contributors are writing a userstyle their first instinct is most likely not to check tips and tricks lol.

sgoudham (commented):
Thoughts on having a `docs/tips-and-tricks.md` instead of a `tips` folder by itself?

uncenter:
I mean that's what we have now, but I don't like it. Some of these examples/tips/triks are getting super long, and the doument as a whole is getting long as we add more tips. I would love an Mdbook of these and the rest of the docs actually if possible, but that might be something for later.

sgoudham:
Alright. I'd rename the `tips` folder to `tips-and-tricks`.

docs/tutorials/writing-a-userstyle.md
sgoudham, line 1: Annoyingly can't suggest what I want but you'll need to update the path in the table so that the images show properly:

https://github.com/catppuccin/userstyles/blob/af16fb8de107a2f689ef2d762c09ac1819a2d27b/docs/tutorials/writing-a-userstyle.md?plain=1#L241-L243

isabelroses (commented):
Re:

> Userstyles labeled with the "🚧" emoji are looking for maintainers, and may
> not work as intended. Contributions are still welcome and encouraged. If you
> would like to become a maintainer, follow the instructions outlined in
> "[Adding yourself as a maintainer](https://github.com/catppuccin/userstyles/blob/main/docs/userstylesyml.md#adding-yourself-as-a-maintainer)."

Instead of updating the link to be the new github link should we not point to the mdbook?

docs/src/README.md
sgoudham, line 1: I feel like it would be "nicer" to symlink the README here, what do you think?

uncenter: I agree sort of; how will we deal with conflicting links? If I link relatively from the symlinked README in the docs directory, it will work on the website, but not on the GitHub readme. And if I link to the website URL from the README, the links aren't really truly accurate on the mdbook page - for example if you were to develop on it and click those links, it would bring you to the deployed version and not the local version because it would be a full URL.

uncenter: Also the "edit this page" link would link to the symlink file on GitHub for the main page, which just contains the text "../../README.md"...

sgoudham: Yeah the only way that it would work is if you replaced the relative GitHub links on the main README with website URLs and avoided clicking those URLs during development. Don't think you can do anything about the "edit this page" link unless you modified the index.hbs ig.

Only reason I mentioned symlinking is that I fear that `userstyles.catppuccin.com` will become the main "entry" point for people to link to and it feels odd to not have all the main README information on the main page of the mdbook.

sgoudham: Also while we're talking about links, it's _probably_ a good idea to run https://github.com/Michael-F-Bryan/mdbook-linkcheck since we're linking to quite a few different places right?

uncenter: Sure. Right now it will fail a ton since all of the userstyles.catppuccin.com links aren't up yet, maybe we add it later.

sgoudham: > Also the "edit this page" link would link to the symlink file on GitHub for the main page, which just contains the text "../../README.md"...

Can you technically flip the symlink so that the primary page is actually from the mdbook and not from the GitHub README? It feels like this book is going to take the spotlight from the GitHub README lol.

docs/src/SUMMARY.md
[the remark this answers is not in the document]
sgoudham: Aside from ctp/mdbook itself, pretty sure ctp/nix is the only one using mdBook which has its own nix'ified CI so probably not too helpful there.

I would take inspiration from ctp/mdbook but probably only run the build on changes to the `docs` directory and `on_workflow_dispatch` for any manual re-runs.

docs/src/tutorials/writing-a-userstyle.md
```diff
@@ -238,8 +226,8 @@ Combining all of the previous steps, we have a working userstyle!
 // vim:ft=less
 ```
 
-| Original                                | Themed                                |
-| --------------------------------------- | ------------------------------------- |
-| ![](./images/example.org-original.webp) | ![](./images/example.org-themed.webp) |
+| Original                                 | Themed                                 |
+| ---------------------------------------- | -------------------------------------- |
+| ![](../images/example.org-original.webp) | ![](../images/example.org-themed.webp) |
```
sgoudham, line 231: This works on GitHub but not on the mdBook. Will most likely have to convert the entire table into HTML.

uncenter: No it's because of Git LFS I believe.

sgoudham: Ah yeah you're right, forgot I was testing this on different machine.

sgoudham (approved):
The CI only gets more cursed but should work as expected... I think.

I'm happy for it to be merged as long as someone goes through https://catppuccin.github.io/userstyles/ to verify it's all good.

uncenter:
Can we get the userstyles.catppuccin.com domain setup? Or else the links in the README don't work and we should update those.

sgoudham:
> Can we get the userstyles.catppuccin.com domain setup? Or else the links in the README don't work and we should update those.

Yeah I was waiting till another approval before setting it up. I know it's been briefly mentioned on discord but I think we should:

1. Use the mdbook-mermaid plugin to make sure the contributing diagram shows properly
2. Symlink `docs/src/CONTRIBUTING` -> `docs/CONTRIBUTING`

NikitaRevenco:
> The Catppuccin project prefers to use the [WebP image format](https://en.wikipedia.org/wiki/WebP) for the asset preview images. We recommend using Google's [cwebp](https://developers.google.com/speed/webp/docs/cwebp) conversion utility to convert images to WebP.

just to let you know, ImageMagick can also convert to webp:

```sh
mogrify -format webp image.png
```

the benefit of using ImageMagick over a dedicated tool would be that users are more likely to have it installed than the specific webp conversion tool

isabelroses (dismissed):

uncenter:
I think I want to get this through on its own as a documentation improvement PR, and then down the road we can look into mdBook or Starlight or whatever is needed for a separate docs site. Didn't want the latter to hold up the former!

isabelroses (dismissed):

docs/SUMMARY.md
sgoudham, line 1: Might be nice to remove this file so it's one less file to update when/if moving the documentation around? Or do you think it's valuable to have even without the mdBook?

uncenter: I think it's a valuable overview atm. If/when we move to a different docs provider that has a nice sidebar overview TOC we can get rid of it imo.

PR #1446, opened 2024-11-13
fix(searxng): default domains, additional tweaks

DokterKaj:
## 🔧 What does this fix? 🔧

- Fixes the default domains not being detected, since `domain()` should not have `https://` in it.
- Fixes additional tweaks not working by explicitly specifying `when (@additions = 1)`

## 🗒 Checklist 🗒

### DIFF
--- a/styles/searxng/catppuccin.user.css
+++ b/styles/searxng/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name SearXNG Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/searxng
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/searxng
-@version 0.3.2
+@version 0.3.3
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/searxng/catppuccin.user.css
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Asearxng
 @description Soothing pastel theme for SearXNG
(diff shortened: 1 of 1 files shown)


styles/searxng/catppuccin.user.css
```diff
@@ -18,7 +18,7 @@
 ==/UserStyle== */
 
 /* Domains picked from https://searx.space/. */
-@-moz-document domain("https://search.bus-hit.me/"), domain("https://search.inetol.net/")
+@-moz-document url-prefix("https://search.bus-hit.me/"), url-prefix("https://search.inetol.net/")
```
isabelroses, line 21: Would this not have worked too?
```suggestion
@-moz-document domain("search.bus-hit.me"), domain("search.inetol.net")
```

DokterKaj: It would, but not all instances are hosted on a specific subdomain. As I mentioned in the PR description, something like `domain("opnxng.com")` would unintentionally affect all subdomains `*.opnxng.com`; additionally, `domain()` wouldn't work for instances not hosted on a subdomain like `darmarit.org/searx`.

Even though the domains I mentioned aren't default domains in the userstyle, when you add an additional domain in Stylus, the URL type starts off as whatever the URL type of the existing domain is, so using `url-prefix()` for the defaults saves a few clicks when adding domains.

uncenter (dismissed):

isabelroses (approved):

PR #1450, opened 2024-11-14
feat(searxng): style guide links

DokterKaj:
## 🔧 What does this fix? 🔧

As the title says, distinguishes links from visited links by making them blue and mauve respectively instead of all being the accent colour.

## 🗒 Checklist 🗒


uncenter:
The default accent is always the same across websites, we don't change it for specific userstyles.

DokterKaj:
In that case it's probably best to hard-code links as blue

Should visited links be hard-coded as mauve or be accented?

uncenter:
Well, according to the style guide visited/followed links would be lavender I believe, not mauve. We haven't made a complete decision on accent links vs style guide links yet, it is likely up to the maintainer to decide which they think looks better for a specific userstyle.

DokterKaj:
Lavender is not quite distinct enough from blue at a glance, so I think I'll still go with mauve like the Brave and DDG userstyles.

DokterKaj:
Not sure where `--color-result-detail-link` is used but it has "link" in the name so best to make it blue anyway

sgoudham:
Just my two cents, I don't think we should accept a contribution that is against the style guide. We're not making a tradeoff to improve some other highlight in this case either.

Also it does make this userstyles inconsistent with the others. As it stands, none of the userstyles support differentiation between links, on hover and visited so we'd probably just have to encourage users to make local changes if they want that at the minute.

In the future, we can probably look towards supporting this with official shades? Curious what @isabelroses and @uncenter think.

DokterKaj:
> Just my two cents, I don't think we should accept a contribution that is against the style guide.

Do the other browser userstyles have to have visited links be changed to lavender then?

> We're not making a tradeoff to improve some other highlight in this case either.

Not quite clear on what this means, sorry

sgoudham:
> Do the other browser userstyles have to have visited links be changed to lavender then?

I believe the style guide wasn't necessarily followed here since lavender itself is offered as an accent colour. I don't think it's feasible to update all the userstyles to that.

We'll probably have to define official shades of colours as part of the style guide so that userstyles can standardise on the amount to lighten/darken the links for hover/visited.

> Not quite clear on what this means, sorry

Sorry ignore me there, what I said wasn't really relevant.

DokterKaj:
> I believe the style guide wasn't necessarily followed here since lavender itself is offered as an accent colour.

This shouldn't be a problem since this PR would make links no longer accented; they'd just be blue and mauve (or blue and lavender, if we must adhere to the style guide) so there shouldn't be any clashing.

> We'll probably have to define official shades of colours as part of the style guide so that userstyles can standardise on the amount to lighten/darken the links for hover/visited.

That would be ideal, but for now this is probably the best way to re-add this functionality of distinguishing visited links.

styles/searxng/catppuccin.user.less
```diff
@@ -44,8 +44,8 @@
     --color-base-font: @text;
     --color-base-background: @base;
     --color-base-background-mobile: @base;
-    --color-url-font: @accent;
-    --color-url-visited-font: @accent;
+    --color-url-font: @blue;
+    --color-url-visited-font: @mauve;
```
uncenter:
```suggestion
    --color-url-visited-font: @lavender;
```

styles/searxng/catppuccin.user.less
```diff
@@ -86,9 +86,9 @@
     --color-result-vim-selected: @surface0;
     --color-result-vim-arrow: @accent;
     --color-result-description-highlight-font: @text;
-    --color-result-link-font: @accent;
-    --color-result-link-font-highlight: @accent;
-    --color-result-link-visited-font: @accent;
+    --color-result-link-font: @blue;
+    --color-result-link-font-highlight: @blue;
+    --color-result-link-visited-font: @mauve;
```
uncenter:
```suggestion
    --color-result-link-visited-font: @lavender;
```

uncenter (changes requested):
Sorry for the delay on this. Updated with main, and I think we should be using lavender instead of mauve for visited links - this matches the "followed links" of the style guide. I will update this in the Google userstyle as well.

uncenter (approved):

ryanccn (commented):
Personally, I disagree with the idea of changing links to the standard style guide link color. IMO in the case of search engines links *are* the primary interactive elements of the user interface and therefore should follow the accent color rather than the usual prose link color. (Otherwise I don't think there would be a whole lot of opportunities to see the configured accent color other than the tab bar.)

uncenter:
> Personally, I disagree with the idea of changing links to the standard style guide link color. IMO in the case of search engines links _are_ the primary interactive elements of the user interface and therefore should follow the accent color rather than the usual prose link color. (Otherwise I don't think there would be a whole lot of opportunities to see the configured accent color other than the tab bar.)

I can agree but then we'll need to document this - specifically define "the primary interactive elements" of an interface and outline the exception to style guide practices. cc @WalkQuackBack

uncenter:
@ryanccn how do you propose we style visited links differently from normal links for a search engine userstyle, assuming the base link is accent? We usually use (recently at least) `lighten(@accent, 5%)` for links on hover, fwiw.

uncenter:
> Lavender is not quite distinct enough from blue at a glance, so I think I'll still go with mauve like the Brave and DDG userstyles.

It's also worth noting this complaint. Not sure if @DokterKaj is still active anymore, but if anyone feels the same they should file an issue on catppuccin/catppuccin and get the style guide changed.

WalkQuackBack:
I'm really sorry for getting to this so late, but I think this is okay to merge for now in parity with the other search engine ports.

WalkQuackBack (approved):

PR #1458, opened 2024-11-15
fix(anilist): change text-bright

AnubisNekhet:
Changes default value of `--color-text-bright`, removes redundant styling


styles/anilist/catppuccin.user.css
```diff
@@ -25,6 +25,16 @@
...
+  // Scrollbar color
+  * {
+    scrollbar-color: @catppuccin[@@lightFlavor][@@accentColor]
+      fade(@catppuccin[@@darkFlavor][@crust], 0.2);
+    &:has(.site-theme-dark) {
+      scrollbar-color: @catppuccin[@@darkFlavor][@@accentColor]
+        fade(@catppuccin[@@darkFlavor][@crust], 0.2);
+    }
+  }
+
```
AnubisNekhet, line 37: I'm not completely sure if this will be fine performance-wise, but it's the only way I could style the scrollbars.

uncenter: Why does it need to be applied to every single element? And why is this not in the #catppuccin mixin?

AnubisNekhet: The userstyle uses the body element to select dark and light flavors, which unfortunately results in the scrollbar (Which is styled using the `*` selector) not being themed. It's the only way I could figure out a workaround, however I'm open to any changes.

styles/anilist/catppuccin.user.css
```diff
@@ -25,6 +25,16 @@
...
 
+  // Scrollbar color
+  * {
+    scrollbar-color: @catppuccin[@@lightFlavor][@@accentColor]
+      fade(@catppuccin[@@darkFlavor][@crust], 0.2);
+    &:has(.site-theme-dark) {
+      scrollbar-color: @catppuccin[@@darkFlavor][@@accentColor]
+        fade(@catppuccin[@@darkFlavor][@crust], 0.2);
+    }
+  }
```
uncenter, line 36: Sorry for taking forever to review this PR. Would you mind updating the PR to modify the Less file, and can we move this section under the #catppuccin mixin? I'm not sure why we need to put it out here and repeat the `@catppuccin[@@lightFlavor][@@accentColor]` stuff.

AnubisNekhet:
I've removed the scrollbar styling entirely since I can't figure out how to make it work without reusing the mixin or using `:has()`. PR seems good to merge now.

uncenter:
Fwiw using :has() is fine?

styles/anilist/catppuccin.user.less
```diff
@@ -102,7 +99,12 @@
     --color-text: #rgbify(@text)[];
     --color-text-light: #rgbify(@subtext1)[];
     --color-text-lighter: #rgbify(@subtext0)[];
-    --color-text-bright: #rgbify(@text)[];
+    & when (@flavor = latte) {
+      --color-text-bright: #rgbify(@crust) [];
+    }
+    & when not (@flavor = latte) {
+      --color-text-bright: #rgbify(@text) [];
+    }
```
uncenter, line 107:
```suggestion
    --color-text-bright: #rgbify(if(@flavor = latte, @crust, @text) [];
```

uncenter (changes requested):

uncenter (approved):

PR #1465, opened 2024-11-20
fix(nixos-manual): name array

isabelroses:
fixes ci failure.
logs: https://github.com/catppuccin/userstyles/actions/runs/11934798547


sgoudham (approved):
If I don't have this yaml refactored away from the array syntax by Christmas, fire me.

PR #1473, opened 2024-11-28
bsky: unthemed colors and some fixes

teidesu:
## 🔧 What does this fix? 🔧

- inappropriate hover in latte
- invisible text in latte
- icon contrast fix in mobile layout
- unthemed skeletons and thread lines

## 🗒 Checklist 🗒

### DIFF
--- a/styles/bsky/catppuccin.user.css
+++ b/styles/bsky/catppuccin.user.css
@@ -2,7 +2,7 @@
 @name Bluesky Social Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/bsky
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/bsky
-@version 0.0.10
+@version 0.1.0
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/bsky/catppuccin.user.css
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Absky
 @description Soothing pastel theme for Bluesky Social
--- a/styles/bsky/catppuccin.user.css
+++ b/styles/bsky/catppuccin.user.css
@@ -2,8 +2,7 @@
 @name Bluesky Social Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/bsky
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/bsky
-@version 0.0.9
-@updateURL https://github.com/catppuccin/userstyles/raw/main/styles/bsky/catppuccin.user.css
+@version 0.0.10
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Absky
 @description Soothing pastel theme for Bluesky Social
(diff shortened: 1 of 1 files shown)


styles/bsky/catppuccin.user.css
```diff
@@ -2,7 +2,7 @@
 @name Bluesky Social Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/bsky
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/bsky
-@version 0.0.9
```
RoootTheFox, line 5: I'd prefer if this was changed to 0.1.0 instead of 0.0.10

styles/bsky/catppuccin.user.css
```diff
@@ -565,6 +578,26 @@
     [style*="background-color: rgb(185, 185, 193)"] {
       background-color: @overlay2 !important;
     }
+    
+    /* skeletons */
+    .r-kdyh1x, .r-cpet4d {
```
RoootTheFox, line 583: are these classes the only way to reliably match skeletons? i've seen these classes change even from small changes, so i'd prefer if there was a more reliable way to match these
(if not then thats okay, we can just update the classes when they (inevitably) break like i do with most of the inline colors already anyways)

RoootTheFox (changes requested):
looks good, does fix the latte issues
i have some very minor issues with it (such as the version being `0.0.10`) but looks good apart from that!

styles/bsky/catppuccin.user.css
teidesu: havent found any really :c

their color is the same as "secondary background color" which is already overridden above but looks pretty bad:
image

these classes set the "border-radius" values on the skeletons which *seem* to only be used on them, so combined with the color this should be safe enough

styles/bsky/catppuccin.user.css
teidesu: i have no idea how versioning is supposed to work in userstyles tbh 🙃

styles/bsky/catppuccin.user.css
RoootTheFox: alright ty !!

RoootTheFox (approved):
looks good!

sgoudham:
I don't use stylus so I don't know if there are any constraints on versioning from that ecosystem but I think it's worth considering again. 

From what I know at the minute, I think datetime versioning makes a lot of sense here (which can also be automatically incremented as part of CI)

uncenter:
> hm, thats fair - i prefer increasing the minor version when patch overflows (9 -> 10)
> 
> maybe consistent versioning scheme guidelines for userstyles are a thing worth considering?

This isn't how we do it. We follow the regular, traditional guidelines for semantic versioning; fixes are patch bumps, features are minor bumps, breaking changes are major bumps.

uncenter:
The commit title here is also incorrect - it doesn't have a commit type. `bsky: unthemed colors and some fixes` needs to be `fix(bsky): misc`.

teidesu:
sorry >_<

PR #1511, opened 2024-12-20
refactor(treewide): cleanup and use rgbify mixin

uncenter:
Adds a new `hslify` mixin to match the functionality of the `rgbify` mixin (extract color values) - not used yet but I did needeto use it for https://github.com/catppuccin/userstyles/pull/1510. Also simplifies the existing `rgbify` mixin and inserts it in places where we should be using it (_technically_ might not be the exact same since the comma vs space separated rgb syntax is weird, will have to check).

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -16,7 +16,7 @@
 
     - [Images and SVGs](./guide/images-and-svgs.md)
     - [Syntax highlighting](./guide/syntax-highlighting.md)
-    - [Extracting RGB values](./guide/rgb-values.md)
+    - [Obtaining raw color values](./guide/raw-color-values.md)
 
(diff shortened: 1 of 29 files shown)


uncenter (commented):
Proton and Snapchat Web are the only two that actually need to be tested, the rest are negligible changes. For these two though the format goes from space separated values to comma separated values, which can break things iirc depending on how the website uses the variable.

uncenter:
Sorry for the pings! Not much to look at here, couldn't figure out how to disable them 😓

NeonGamerBot-QK (approved):
lgtm - ollama

trinkey (approved):

isabelroses (approved):

Nyatalieeee (approved):

PR #1516, opened 2024-12-24
docs(README.md): remove mention of adding yourself as a maintainer

uncenter:
We don't want to encourage new contributors to immediately become maintainers. We can suggest the idea of becoming a maintainer to active contributors when the time is right imo.


isabelroses (approved):

isabelroses:
perhaps we should follow this up with a rough guide on the steps to taking over a style?

uncenter:
Eh I think that can be explained in Discord or as we go, its hard to document that as it isn't some sort of exact procedure.

PR #1525, opened 2024-12-25
fix(hacker-news): header, logo, and story ranking color

ashish0kumar:
## 🔧 What does this fix? 🔧

Fixed the header and logo to use the accent color and corrected the story ranking number color. 
It appeared as expected before, but I believe recent changes to the site caused it to break.

| **Before**                    | **After**                 |
| ----------------------------- | ------------------------- |
| image | image |

## 🗒 Checklist 🗒

### DIFF
--- a/styles/hacker-news/catppuccin.user.css
+++ b/styles/hacker-news/catppuccin.user.css
@@ -151,6 +151,11 @@
       color: @lavender;
     }
 
+    font[color="#005a00"],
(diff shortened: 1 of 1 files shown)


uncenter:
Wild guess but I think the changes to the site might be time limited? It looks Christmas-themed so I would expect it to go away in a few days and I don't want to have to revert this at that point.

PR #1550, opened 2025-01-01
fix(hackage): make headings text

uncenter:
## 🔧 What does this fix? 🔧

| Before | After |
| --- | --- |
| CleanShot 2024-12-31 at 20 38 08 | CleanShot 2024-12-31 at 20 37 26 |
## 🗒 Checklist 🗒

### DIFF
--- a/styles/hackage/catppuccin.user.less
+++ b/styles/hackage/catppuccin.user.less
@@ -110,6 +110,7 @@
     .section5,
     .section6 {
       color: @text;
+      filter: none;
     }
 
     .collapser::before,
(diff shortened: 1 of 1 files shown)


styles/hackage/catppuccin.user.less
```diff
@@ -109,8 +109,7 @@
     .section4,
     .section5,
     .section6 {
-      color: overlay(@accent, @surface0);
-      filter: none;
```
jn-sena, line 113: I think the `filter: none;` should stay, as they use brightness filters in some places, which leads to this [here](https://hackage.haskell.org):
screenshot

jn-sena (changes requested):
I like it! I was thinking about maybe redesigning the colors so that the headers are accent and the links are maybe rosewater or blue or something, but this is good too. We can merge it after fixing the filter unless it was intentional.

jn-sena (approved):
Thanks! ❤️

PR #1557, opened 2025-01-04
feat(pluralkit-dash): init

waterlilly-lilly:
## 🎉 Theme for [PluralKit Dashboard](https://dash.pluralkit.me) 🎉

PluralKit Dashboard is a Web-based dashboard for the [PluralKit](https://pluralkit.me) Discord bot.

Screenshot 2025-01-04 at 16-15-11 PluralKit profile
Screenshot 2025-01-04 at 16-14-58 PluralKit dash

## 💬 Additional Comments 💬

- I couldn't include any screenshots showing the member or group pages due to licensing concerns with the profile pictures I use. Since these are the main part that a user will interact with, it might be worth finding a solution like creating an example system where this isn't a concern.
- I had to use a bit of a kludge to color a few UI element SVGs with hardcoded colors (involving the `e()` function from lesscss)
- Similar story with the status page, where I had to use an attribute selector on the `style` attribute for the shard class in order to style the colors.
- I'm not totally sure what the standard practice for borders is, so I used color offsets from the element background color. This could be pretty easily changed, though.
- When I was editing it in the builtin Stylus editor, it showed an error on line 25 ("unknown word"). It didn't seem to affect anything, so I'm not sure if it's a problem or not.
- I could probably theme the main pluralkit.me page with relative ease, though the average PluralKit user interacts with the dashboard much more.
- I don't think this is too niche, considering that the bot is in nearly a million servers, but I'm not sure what the criteria for nicheness is.

## 🗒 Checklist 🗒

- [X] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [X] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [X] I have ensured that the new directory is in **lower-kebab-case**.
  - [X] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [X] I have made sure to update the [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml) file with information about the new userstyle.
- [X] I have included the following files:
  - [X] `catppuccin.user.less` - all the CSS for the userstyle, based on the template.
  - [X] `preview.webp` - composite image of all four individual flavor screenshots (taken with the default accent color of mauve) stitched together, generated via [Catwalk](https://github.com/catppuccin/catwalk).


scripts/userstyles.yml
```diff
@@ -610,10 +613,11 @@ userstyles:
...
     readme:
-      app-link: [
-        https://github.com/libreddit/libreddit,
-        https://github.com/redlib-org/redlib,
-      ]
+      app-link:
+        [
+          https://github.com/libreddit/libreddit,
+          https://github.com/redlib-org/redlib,
+        ]
```
uncenter, line 620: I don't think this should happen if the Deno formatter is used. This might be from Prettier, which is no longer used.

scripts/userstyles.yml
```diff
@@ -840,6 +844,13 @@ userstyles:
     readme:
       app-link: https://planetminecraft.com
     current-maintainers: [*nyameliaaaa]
+  pluralkit-dash:
+    name: Pluralkit Dashboard
```
uncenter: I think this should just be called `PluralKit`? Why specifically "Dashboard"?

styles/pluralkit-dash/frappe.webp
uncenter, line 1: We no longer do individual accent screenshots, only the `preview.webp` is necessary.

uncenter (changes requested):

styles/pluralkit-dash/catppuccin.user.less
```diff
@@ -129,7 +129,7 @@
       color: @text;
     }
     .btn:focus {
-      box-shadow: 0 0 0 0.25rem fadeout(lighten(@accent, 10%), 75%);
+      box-shadow: 0 0 0 0.25rem fade(lighten(@accent, 10%), 75%);
```
uncenter, line 132: This is incorrect, the two are not interchangable.

image

If you run `deno task lint:fix` instead of manually changing this, it will correctly fix it for you fyi.

scripts/userstyles.yml
waterlilly-lilly: Initially I was just styling the dashboard (at `dash.pluralkit.me`), but I can also probably style the main website too.

waterlilly-lilly:
recreating PR under feat(pluralkit) instead of feat(pluralkit-dash)

PR #1566, opened 2025-01-06
fix(twitch): dashboard fixes

mxgic1337:
## 🔧 What does this fix? 🔧

- Fixes achievement border color
- Fixes clips length text color
- Fixes collection background color
- Fixes _Account suspended or deactivated_ status banner background color
- Themes stream manager quick actions
- Themes activity feed icons
- Themes chat message highlights
- Themes chat mentions

## 🗒 Checklist 🗒

### DIFF
--- a/styles/twitch/catppuccin.user.less
+++ b/styles/twitch/catppuccin.user.less
@@ -1070,6 +1070,13 @@
         border-color: @blue !important;
       }
     }
+
+    /* "Account suspended or deactivated" status banner */
+
+    .viewer-card-header__account-status-banner {
+      color: @text;
+      background-color: @mantle !important;
+    }
   }
 }
 
--- a/styles/twitch/catppuccin.user.less
+++ b/styles/twitch/catppuccin.user.less
@@ -1076,6 +1076,19 @@
       color: @text;
       background-color: @mantle !important;
     }
+
+    /* Chat mentions */
+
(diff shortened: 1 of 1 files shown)


styles/twitch/catppuccin.user.less
```diff
@@ -629,6 +629,7 @@
       .channel-root,
       .channel-info-content,
       .chat-author__display-name,
+      .seventv-chat-user-username span,
```
uncenter, line 632: This should go under the seventv userstyle.

mxgic1337: I think it should stay here, it's more of an issue with the Twitch userstyle. People that use the Twitch userstyle only with the 7TV extension installed will still see this issue.

uncenter: Shouldn't it be in https://github.com/catppuccin/userstyles/pull/1552?

mxgic1337: No, It's an issue in the Twitch userstyle.

uncenter: It has the word "seventv" in it. I don't understand how this *isn't* related to the 7TV extension? Would you not only see this in the HTML if you are using seventv?

uncenter (approved):

PR #1569, opened 2025-01-09
fix(chatgpt): limit alert bg and middle pricing panel

ashish0kumar:
## 🔧 What does this fix? 🔧

Fixes the backgrounds of limit alert and middle pricing panel (which is on the upgrade plan page)

|**Before**|**After**|
|----------|---------|
|image|image|
|image|image|

## 🗒 Checklist 🗒

### DIFF
--- a/styles/chatgpt/catppuccin.user.less
+++ b/styles/chatgpt/catppuccin.user.less
@@ -386,7 +386,7 @@
     }
 
     .border-green-600 {
-      border-color: @green;
+      border-color: @accent;
     }
 
     .border-orange-400 {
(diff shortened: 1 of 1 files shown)


styles/chatgpt/catppuccin.user.less
```diff
@@ -613,10 +609,6 @@
       background-color: fade(lighten(@green, 10%), 10%);
     }
 
-    .bg-green-600 {
-      background-color: @green;
-    }
```
uncenter, line 618: This still needs to be there to theme the background of the middle panel, something like `fade(@accent, 20%)` maybe.

PR #1577, opened 2025-01-13
docs: improve unmaintained notes & emojis in READMEs

uncenter:
Importantly this removes the note about adding yourself as a maintainer - I don't think we want to encourage this. I also decided to change the emoji from the foreboding caution gate emoji to this more fitting broken(?) heart emoji, in line with the current/past maintainers emojis, because userstyles that don't have maintainers can still very well be perfectly fine/working correctly despite that.


isabelroses (approved):
I think i still prefer the the old emoji, but the mending heart is a nice touch

PR #1605, opened 2025-01-28
fix(chess.com): board colors

Aimnos:
## 🔧 What does this fix? 🔧
Created new variables for the colors used by the board's light and dark cells and used them to fix the board colors, as well as the coordinates' colors, for dark themes.

## 🗒 Checklist 🗒


Aimnos:
> Before and after screenshots would be appreciated to better understand the original issue.

Absolutely!
The colors of the squares in the userstyle dark flavors are opposite those in the original board, as shown below.

Chess.com default board (top left cell is light):
image

Catppuccin userstyle Latte board (top left cell is light, although the pieces are the wrong colors):
image

Catppuccin userstyle Mocha board before fix (top left cell is dark):
image

Catppuccin userstyle Mocha board after fix (top left cell is light):
image

uncenter:
Does this also fix the piece colors on latte being incorrect as you mentioned?

Aimnos:
> Does this also fix the piece colors on latte being incorrect as you mentioned?

It does not. I didn't even notice that they were wrong until I took the screenshots to post here. Should I open a new issue with that?

styles/chess.com/catppuccin.user.less
```diff
@@ -56,6 +56,9 @@
     @accent: @catppuccin[@@flavor][@@accentColor];
 
     @highlight: @catppuccin[@@flavor][@@highlightColor];
+      
+    @light-cell: if(@flavor = latte, @catppuccin[@@flavor][@surface0], @catppuccin[@@flavor][@@accentColor]);
+    @dark-cell: if(@flavor = latte, @catppuccin[@@flavor][@@accentColor], @catppuccin[@@flavor][@surface0]);
```
uncenter, line 61:
```suggestion
    @light-cell: if(@flavor = latte, @surface0, @accent);
    @dark-cell: if(@flavor = latte, @accent, @surface0);
```
Is there a reason we aren't just doing this?

uncenter (changes requested):

styles/chess.com/catppuccin.user.less
Aimnos: Oh... yeah. My bad. I'm not very experienced with CSS, but it should have been obvious to me that I didn't need to reuse the definition for those variables.

styles/chess.com/catppuccin.user.less
```diff
@@ -56,6 +56,9 @@
     @accent: @catppuccin[@@flavor][@@accentColor];
 
     @highlight: @catppuccin[@@flavor][@@highlightColor];
+      
+    @light-cell: if(@flavor = latte, @surface0, @accent);
+    @dark-cell: if(@flavor = latte, @accent, @surface0);
```
uncenter, line 61:
```suggestion
```
Let's move these variables down below to where we actually use them.

styles/chess.com/catppuccin.user.less
```diff
@@ -609,7 +612,7 @@
 
     /* Board */
```
uncenter, line 613:
```suggestion
    /* Board */
    @light-cell: if(@flavor = latte, @surface0, @accent);
    @dark-cell: if(@flavor = latte, @accent, @surface0);

```

uncenter (approved):
Thank you! ❤️

PR #1612, opened 2025-01-30
fix(lobste.rs): match changed applicators & variables

uncenter:
Change happened in https://github.com/lobsters/lobsters/commit/2de4eee079e4b2904b6bcd17b78a78db2129f990, I don't think there is another way to detect the user-configured theme otherwise.


steinuil (approved):
LGTM! I took a look at the changes on the lobste.rs repo, looks like checking the URL of the loaded stylesheets is indeed the only way to detect the colorscheme now. Thanks for updating it!

PR #1614, opened 2025-01-30
fix(wikipedia): remove default anchor element color

### LINKED ISSUE #1613: wikipedia / wiktionary: colored links are indistinguishable

lbl:wikipedia

### Describe your problem.

Wikipedia and Wiktionary both use differently colored links to highlight certain meanings - blue for a working link, and red for a link to a page that doesn't exist. The English Wiktionary also has a toggle (for registered users only) that makes certain links orange (when the page exists, but doesn't have an entry for the language being linked to). As I work with these platforms a lot, it is important to me to be able to distinguish these link types. The current userstyle makes all of these links the same color, making it unusable for me.

I would like to see the different link colors be mapped to their closest match in the catppuccin color palette. The "blue" links could keep the primary accent color, but that may lead to confusion if it is set to a color that overlaps with the others. So maybe it should be forced to always be blue. Not sure

### Attach screenshots.

Section of a Wiktionary page without the style enabled. Various links are shown, some of which are blue, orange, and red.
Screenshot of a Wiktionary page with the style enabled. All the links are now the same color.

Firefox for Android 134.0.2

### Any additional comments?

_No response_

uncenter:
## 🔧 What does this fix? 🔧

Closes #1613. Leaving as a draft for now since I'm pretty sure I had this set for a reason when I rewrote it (but I can't find what the issue was now).

## 🗒 Checklist 🗒

### DIFF
--- a/styles/wikipedia/catppuccin.user.less
+++ b/styles/wikipedia/catppuccin.user.less
@@ -232,8 +232,6 @@
     }
 
     a {
-      color: @accent;
-
       &:focus {
         outline-color: @accent;
       }


uncenter:
If it breaks something it breaks something, can fix later with a more specific selector ig.

PR #1410, opened 2024-10-28
fix(google-gemini): misc

Dandraghas:
## 🔧 What does this fix? 🔧

Closes #1406

Fixes
* Outlined text field
* Google logo
* Banner
* Expandable menu
* Checkbox
* Activity icon
* Accent-filled button
* Unused variables

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [contributing guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md).
- [x] I have updated the version appropriately in the `==UserStyle==` header of the `catppuccin.user.css` file.

Issue #1406, opened 2024-10-27
Gemini - Black Shadow appearing under prompt


mambucodev:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:google-gemini

### Describe your problem.

A black shadow appears underneath the prompt element.

Mantainer: @Dandraghas 

### Attach screenshots.

![image](https://github.com/user-attachments/assets/6861e2da-f603-4141-9d41-a898602bb71a)

### What browser(s) are you seeing the problem on?

Firefox v131.0.3

PR #1624, opened 2025-02-05
fix(google-gemini): remove weird shadows

Lunchb0ne:
## 🔧 What does this fix? 🔧

Fixes some weird lingering shadow issues on the Gemini Website. Seems similar to #1406 that was fixed in #1410. Looks like the CSS for the site was updated again and that broke it.

Before the fix:
image

Here's after the fix:
image

## 🗒 Checklist 🗒


Dandraghas:
> ## 🔧 What does this fix? 🔧
> 
> Fixes some weird lingering shadow issues on the Gemini Website. Seems similar to #1406 that was fixed in #1410. Looks like the CSS for the site was updated again and that broke it.
> 
> Before the fix: image
> 
> Here's after the fix: image
> ## 🗒 Checklist 🗒
> 
>     * [x]  I have read and followed Catppuccin's [contributing guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md).

I can't reproduce your issue:
image
image
input-container `background` seems to already follow `--gem-sys-color--surface--rgb`, so your fixes shouldn't be necessary. does it still happens?

Lunchb0ne:
Could it be that because I have Gemini Advanced, the page is slightly different?

uncenter:
I can't reproduce either, are you still experiencing the issue @Lunchb0ne? If so it is likely as you said, caused by the difference in HTML of basic vs advanced plans, and I'll review for a potential merge.

Lunchb0ne:
Yeah, still experiencing it..

uncenter (approved):

PR #1642, opened 2025-03-06
fix(crates.io): update css variables

mxgic1337:
## 🔧 What does this fix? 🔧

Seems like the names of CSS variables changed and some stuff got broken. This pull request updates them.

## 🗒 Checklist 🗒


uncenter:
This userstyle is the bane of my existence... didn't realize how often they updated their variables (I think they might be randomly generated on every build, non-deterministically). I wonder if we can subtly contribute upstream to improve this flow -_-

uncenter (approved):

PR #1644, opened 2025-03-07
fix(pypi): unstyled tables

ashish0kumar:
## 🔧 What does this fix? 🔧

Fixes unstyled tables

## Screenshots

|**Before**|**After**|
|---|---|
|image|image|

## 🗒 Checklist 🗒

### DIFF
--- a/styles/pypi/catppuccin.user.less
+++ b/styles/pypi/catppuccin.user.less
@@ -136,12 +136,14 @@
       color: @subtext0 !important;
     }
 
-    .project-description table,
-    .project-description table tr,
-    .project-description table tr th,
-    .project-description table tr td {
-      background-color: @base !important;
-      border-color: @surface2 !important;
+    .project-description table {
+      &,
+      tr,
+      th,
+      td {
+        background-color: @base !important;
+        border-color: @surface0 !important;
+      }
     }
 
     .banner,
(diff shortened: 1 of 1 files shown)


styles/pypi/catppuccin.user.less
```diff
@@ -136,6 +136,14 @@
       color: @subtext0 !important;
     }
 
+    .project-description table,
+    .project-description table tr,
+    .project-description table tr th,
+    .project-description table tr td {
+      background-color: @base !important;
+      border-color: @surface2 !important;
```
trinkey, line 144: maybe use `@surface0` to be consistent with the other borders?

also the selector could be rewritten as:
```css
.project-description table {
  &,
  tr,
  th,
  td {
    ...
  }
}
```

trinkey (changes requested):

uncenter:
Please also provide a URL to reproduce the issue, otherwise I cannot test the issue before/after.

trinkey:
https://pypi.org/project/yt-dlp/

it happens on projects that have a table in the readme

trinkey (approved):
lgtm. version is automatically done now, right @uncenter?

uncenter:
> lgtm. version is automatically done now, right @uncenter?

Yep it is!

uncenter (changes requested):
If we look at the original styling for the table, the color for every other row is slightly altered:

CleanShot 2025-03-07 at 15 19 52

Can we do the same here?

EDIT: I had a guess that this would be merged before I typed this out ahah (no worries, my fault). Could one of you do this in a follow up PR?

PR #1651, opened 2025-03-11
fix(chess.com): arrow and square highlights

### LINKED ISSUE #1606: Chess.com arrow/square highlight colors are wrong

lbl:chess.com

### Describe your problem.

When highlighting a square by pressing the right mouse button on the board or when drawing an arrow by dragging the mouse while holding the right mouse button, you can change the color of the arrow/square highlight by simultaneously holding one of the modifier keys (Ctrl, Shift or Alt).
The colors available by default are red, blue, green and yellow. When using the catppuccin userstyle, all arrows become grey and all highlighted squares become a single color dependent on the theme's highlight color.

### Attach screenshots.

Default behavior:
Image

Behavior when using the catppuccin userstyle with a maroon highlight:

Image

1.74.48 Chromium: 132.0.6834.83 (Official Build) (64-bit)

### Any additional comments?

_No response_

nuexq:
## 🔧 What does this fix? 🔧

Fixes #1606.

image

- [x] arrow 4 variant color fix
- [x] highlight 4 variant color fix

## 🗒 Checklist 🗒

### DIFF
--- a/styles/chess.com/catppuccin.user.less
+++ b/styles/chess.com/catppuccin.user.less
@@ -447,16 +447,16 @@
 
     .highlight {
       &[style*="rgb(235, 97, 80)"] {
-        background: @highlight1 !important;
+        background: fade(@highlight1, 80%) !important;
       }
       &[style*="rgb(172, 206, 89)"] {
-        background: @highlight2 !important;
+        background: fade(@highlight2, 80%) !important;
       }
       &[style*="rgb(255, 170, 0)"] {
-        background: @highlight3 !important;
+        background: fade(@highlight3, 80%) !important;
       }
       &[style*="rgb(82, 176, 220)"] {
-        background: @highlight4 !important;
+        background: fade(@highlight4, 80%) !important;
       }
       opacity: 1 !important;
     }
--- a/styles/chess.com/catppuccin.user.less
+++ b/styles/chess.com/catppuccin.user.less
@@ -481,7 +481,18 @@
     }
 
     .arrow {
-      fill: @overlay0 !important;
+      &[style*="rgba(248, 85, 63, 0.8)"] {
+        fill: @red !important;
+      }
+      &[style*="rgba(255, 170, 0, 0.8)"] {
+        fill: @peach !important;
+      }
+      &[style*="rgba(72, 193, 249, 0.8)"] {
+        fill: @blue !important;
+      }
+      &[style*="rgba(159, 207, 63, 0.8)"] {
+        fill: @green !important;
+      }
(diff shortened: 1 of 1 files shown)


nuexq:
I've noticed that there is a `@var` for selecting highlight color, but there is 4 variant of them in the website but the theme make you select just one?

uncenter:
That is correct, see https://github.com/catppuccin/userstyles/issues/1606. Not intentional, I just wasn't aware. If you would like to, I would love if you could implement those additional variables and style them when appropriate to close that issue!

nuexq:
do I need to make them configurable or just using the red/peach/blue/green like the arrows and remove the `@var` for highlightColor ?

uncenter:
I originally made it configurable so that the highlight colors wouldn't clash with your chosen accent color (if you had a blue accent, those arrows/highlights would be hard to see!). I think we should continue the same here, and maybe call them highlightColor1/highlightColor2/highlightColor3/highlightColor4?

styles/chess.com/catppuccin.user.less
```diff
@@ -15,6 +15,10 @@
 @var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
 
 @var select highlightColor "Highlight" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve", "red:Red", "maroon:Maroon*", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
```
nuexq, line 17: I did leave the original highlightColor because it's used on `--theme-board-style-highlight-color` and I don't know where that variable used.

nuexq:
The color swap between `Shift + Right Click` and `Right Click` in the original theme is due to a bug on Chess.com. This issue causes the red and peach square be reversed compared to how it should look based on it's arrow.

styles/chess.com/catppuccin.user.less
uncenter: I think we can remove the original highlightColor and change the reference to `@highlight` for `--theme-board-style-highlight-color` as you mentioned to use `@highlight1` instead?

PR #1658, opened 2025-03-14
refactor(scripts): pull in separate `categories.schema.json`

sgoudham:
Draft until I figure out how to get editor (vscode, neovim, etc) validation to
work as well as ajv itself.

I imagine it's due to some misunderstanding of the semantics of `$ref` and `$id`
on my part.


scripts/userstyles.schema.json
```diff
@@ -183,58 +207,20 @@
...
     },
     "categories": {
       "$id": "#categories",
       "type": "array",
+      "description": "The categories that fit the port the most, the first category is the primary category which the port will be listed under on the README.",
       "minItems": 1,
       "maxItems": 3,
       "items": {
-        "$ref": "#/$defs/category"
+        "$ref": "https://raw.githubusercontent.com/catppuccin/catppuccin/refs/heads/main/resources/categories.schema.json#/$defs/category"
```
sgoudham, line 222: I _think_ the only quirk of this is that while it's technically pointing to a remote file, this is the `$id` of the categories.schema.json and therefore, it'll be linked to the hash of the repo in `deno.json`, not actually pointing towards `main`. 

Winston has nerdsniped me a little into looking at https://github.com/tweag/nickel so I might revisit this entire thing at a later point.

uncenter (approved):

PR #1663, opened 2025-03-14
fix(wikipedia): page previews

### LINKED ISSUE #1596: Wikipedia Page Preview is colored incorrectly

lbl:wikipedia

### Describe your problem.

A part of the "Page Preview" popup on Wikipedia is colored incorrectly (see screenshots).

Image
The gradient here in the `::after` should go from transparent to the page background color (`--background-color-base` I think) instead of white (`#ffffff`).

To also fix this for right-to-left languages, the same should be done for `.mwe-popups-extract[dir='rtl']::after` with a gradient to left.
If it helps, the default of a right-to-left written article is this:
Image

### Attach screenshots.

Image

Image

Chromium 131.0.6778.264

### Any additional comments?

hope this helps 😀

Mouadhbendjedidi:
## 🔧 What does this fix? 🔧

Closes #1596

## 🗒 Checklist 🗒


Mouadhbendjedidi:
did i forget something?

styles/wikipedia/catppuccin.user.less
```diff
@@ -372,6 +372,14 @@
     .languagesettings-menu .menu-section.active {
       background-color: @surface1;
     }
+    .mwe-popups .mwe-popups-extract[dir='ltr']::after {
+		  right: 0;
```
uncenter, line 376: We shouldn't need to set the `right` or `left` properties here.

Mouadhbendjedidi: so i just need to delete them?

uncenter: Yep, thanks.

uncenter (approved):

PR #1665, opened 2025-03-21
feat(hardcover): init

Papweer:
## 🎉 Theme for Hardcover 🎉

Hardcover is a website to track, discover and share books. [hardcover.app](http://hardcover.app)

## 💬 Additional Comments 💬

I couldn't figure out how to theme the tool tips that appear when you hover over a book icon. They don't appear with pusdoclasses and disapear when freezing the page.

## 🗒 Checklist 🗒

- [X] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [X] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [X] I have ensured that the new directory is in **lower-kebab-case**.
  - [X] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [X] I have made sure to update the [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml) file with information about the new userstyle.
- [X] I have included the following files:
  - [X] `catppuccin.user.less` - all the CSS for the userstyle, based on the template.
  - [X] `preview.webp` - composite image of all four individual flavor screenshots (taken with the default accent color of mauve) stitched together, generated via [Catwalk](https://github.com/catppuccin/catwalk).


Papweer:
> Have you tried https://github.com/catppuccin/userstyles/blob/main/docs/tips-and-tricks/inspect-hard-to-grab-elements.md?

The tooltip just seems to not exist when I tried to pick them when frozen (or at any other time)

scripts/userstyles.yml
```diff
@@ -429,6 +431,13 @@ userstyles:
     readme:
       app-link: https://news.ycombinator.com
     current-maintainers: [*lucasmelin]
+  hardcover:
+    name: Hardcover
+    categories: [library, social_networking]
```
uncenter, line 436:
```suggestion
    categories: [productivity, social_networking]
```
The library category is unfortunately not for things relating to actual book libraries, it is for software libraries 😅

scripts/userstyles.yml
```diff
@@ -429,6 +431,13 @@ userstyles:
...
+    color: mauve
```
uncenter, line 437:
```suggestion
    color: lavender
```
Feels more appropriate.

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    input,
+    textarea {
+      &::placeholder {
+        color: @subtext0 !important;
+      }
+    }
+
+    :root, body, html {
```
uncenter, line 68:
```suggestion
    &, body {
```

`&` refers to the parent selector, which in this case is already `:root` (since we apply this #catppuccin mixin within a :root selector on lines 19-24 above). And `:root` is another way of saying the `html` element.

WalkQuackBack:
> > Have you tried https://github.com/catppuccin/userstyles/blob/main/docs/tips-and-tricks/inspect-hard-to-grab-elements.md?
> 
> The tooltip just seems to not exist when I tried to pick them when frozen (or at any other time)

By not exist does it entirely disappear from the DOM tree? It should still be there in the debugger view, you just might not be able to pick it using the visual picker. If you know keywords you can manually search the DOM for the element while it's frozen.

uncenter:
> The tooltip just seems to not exist when I tried to pick them when frozen (or at any other time)

I think your issue is timing. You'll need to run the snippet in the console, and then move your mouse over the page and trigger the action (e.g. hover over the thing that causes the tooltip to show up) in the three seconds before the page freezes. Then with the page frozen you should be able to still see the tooltip you triggered and you can inspect it and check it out.

Papweer:
Fixed the tooltips. Turns out the element for them was at the very bottom of the html.

scripts/userstyles.yml
```diff
@@ -333,6 +334,13 @@ userstyles:
     icon: ycombinator
     color: peach
     current-maintainers: [*lucasmelin]
+  hardcover:
+    name: Hardcover
+    categories: [productivity, social_networking]
+    color: lavender
+    readme:
+      app-link: https://hardcover.app
```
uncenter, line 342:
```suggestion
    name: Hardcover
    link: https://hardcover.app
    categories: [productivity, social_networking]
    color: lavender
```

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,362 @@
...
+        color: @subtext0 !important;
+      }
+    }
+
+    &, body {
+      background-color: @base !important;
+    }
+
+    // Base Elements
+    p, div, a, span, button, h1, h2, h3, strong {
```
uncenter, line 73: We should target the Tailwind classes or variables instead of individual elements. In this case maybe `.text-white`.

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,362 @@
...
+      color: @text;
+    }
+
+    // General
+      // Backgrounds
+    .bg-white {
+      background-color: @crust;
+      border-color: @crust;
```
uncenter, line 81: Target whatever other Tailwind class is setting the border color, since `bg-white` only sets the background.

uncenter: Same goes for a bunch of these classnames below.

uncenter (changes requested):

Papweer:
Should be good now, style is much cleaner now since I'm using mixins and as far as I can see everything is themed.

Papweer (commented):
Should be ready to review now

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,336 @@
...
+      // Target all text elements in tooltip
+      span, a, h1, h2, div {
+        color: @base !important;
+      }
+    }
+    
+    // Transparency
+    .bg-opacity-70 {
+      opacity: 0.8 !important;
+    }
```
uncenter: This seems like something that they would already have set?

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,336 @@
...
+    .generateColorStyles(~"amber", @yellow);
+    .generateColorStyles(~"blue", @blue);
+    .generateColorStyles(~"orange", @peach);
+
+    // Border Colors
+    @borderMantleClasses: ~".lg\:border-secondary, .hover\:bg-gray-100:hover, .border-gray-100, .dark\:border-gray-700, .border-secondary, .border-tertiary, .dark\:border-secondary";
+    @borderSurface0Classes: ~".border-gray-200, .border-gray-300, .border-gray-400, .border-muted";
+
+    @{borderMantleClasses} { border-color: @mantle; }
+    @{borderSurface0Classes} { border-color: @surface0; }
```
uncenter: Is there a reason for these being in strings instead of just being selectors?

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,336 @@
...
+    @{bgMantleClasses} { background-color: @mantle; }
+    @{bgSurface0Classes} { background-color: @surface0; }
+    @{bgSurface1Classes} { background-color: @surface1; }
+
+    // Accent + Colored Backgrounds
+    @textAccentClasses: ~".text-accent, .text-primary, .text-indigo-100, .text-indigo-300, .text-indigo-400, .text-indigo-500, .text-indigo-600, .text-indigo-700, .text-indigo-800, .hover\:text-accent:hover";
+    @borderAccentClasses: ~".border-accent, .border-primary, .border-indigo-50, .border-indigo-100, .border-indigo-200, .border-indigo-300, .border-indigo-400, .border-indigo-500, .border-indigo-600, .border-indigo-700, .hover\:border-accent:hover, .hover\:border-primary:hover, .hover\:border-indigo-50:hover, .hover\:border-indigo-100:hover, .hover\:border-indigo-200:hover, .hover\:border-indigo-300:hover, .hover\:border-indigo-900:hover";
+    
+    @{borderAccentClasses} { border-color: @accent !important; }
+    @{textAccentClasses} { color: @accent !important; }
```
uncenter: Same here, seems like we can just use these as selectors instead of inline strings.

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,336 @@
...
+      }
+
+      // Text classes
+      [class*=" text-@{colorName}"], [class^="text-@{colorName}"] {
+        color: @colorValue;
+      }
+      // Fill
+      [class*=" fill-@{colorName}"], [class^="fill-@{colorName}"] {
+        fill: @colorValue;
+      }
```
uncenter: Why do we need the space after the opening quote? If you don't do that the second part with ^= is seemingly unnecessary?
```suggestion
      [class*="bg-@{colorName}"] {
        .coloredElement(@colorValue);
      }

      // Border classes
      [class*="border-@{colorName}"] {
        border-color: @colorValue;
      }

      // Text classes
      [class*="text-@{colorName}"] {
        color: @colorValue;
      }
      // Fill
      [class*="fill-@{colorName}"] {
        fill: @colorValue;
      }
```

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,336 @@
...
+      fill: @textColor !important;
+      --tw-ring-color: @bgColor;
+
+      [class*="text-"], [class*="fill-"] {
+        color: @textColor !important;
+        fill: @textColor !important;
+      }
+    }
+
+    .generateColorStyles(@colorName, @colorValue) {
```
uncenter:
```suggestion
    .tailwindColorUtilities(@colorName, @colorValue) {
```

I would rename it to something like this. Don't forget to update uses of it down below.

uncenter (changes requested):

uncenter (changes requested):
This bar at the bottom is unthemed:

CleanShot 2025-07-27 at 13 48 11

The background of a list like this is unthemed (https://hardcover.app/@hardcover/lists/the-esquire-75-best-sci-fi-books-of-all-time?order=list_books_position&direction=asc&start=0&limit=100&view=shelf):

CleanShot 2025-07-27 at 13 48 15

Seems to be a black border around this seaerch bar at the top?

CleanShot 2025-07-27 at 13 48 21

styles/hardcover/catppuccin.user.less
Papweer: `[class*="bg-"]` would also match classes starting with `dark:`. The space makes sure that the class is not prefixed by anything.

uncenter: Can we leave a comment like `/* Spaces added before classnames and additional =^ selector used to avoid matching classes starting with "dark:". */` then?

styles/hardcover/catppuccin.user.less
Papweer: Renamed

styles/hardcover/catppuccin.user.less
Papweer: Was a remnant of some stuff I was trying, removed now.

styles/hardcover/catppuccin.user.less
Papweer: Same reason as above, fixed

Papweer:
Should be good now?

uncenter:
No need to update this branch it pings me everytime and there is no reason to be up to date with main all the time.

uncenter:
The text of these tab headers in the login/signup modal is unthemed on hover (seems like setting `--color-white` to `text` might fix it):

CleanShot 2025-08-21 at 15 49 10

This first account steps page has some issues:
CleanShot 2025-08-21 at 15 51 21

1) These bars at the top left are unthemed, are themed like so:

```css
.bg-indigo-500 {
	background-color: var(--color-indigo-500);
}
```

They should be distinct from the accent color, slightly darker or lighter maybe.

2) Placeholder text in the form inputs is unthemed, is themed like:

```css
.text-foreground\/50 {
	color: color-mix(in oklab,var(--foreground)50%,transparent);
}
```

You could set `--foreground` but I would set the whole `.text-foreground\/50` class to `@subtext0` to match how we do placeholders elsewhere instead of letting their transparency attempt it.

Unthemed stuff on the edit profile page:

CleanShot 2025-08-21 at 15 55 13

The edit profile button itself shouldn't be yellow on hover/active I think:

CleanShot 2025-08-21 at 15 57 07


And on the account settings privacy page:

CleanShot 2025-08-21 at 15 55 54

Seems like the border colors are a bit too muted, especially compared to the unthemed site:

CleanShot 2025-08-21 at 15 58 37

CleanShot 2025-08-21 at 15 58 45

The bar chart primary color for ratings should be yellow, not peach, to match the star color:

CleanShot 2025-08-21 at 15 59 45

Unlike the unthemed site, hovering over the icons in the bottom bar doesn't show any indication of hover state.

CleanShot 2025-08-21 at 16 01 12
CleanShot 2025-08-21 at 16 01 28

Papweer:
Fixed all the issues you mentioned and some more I found. Code is cleaner now, should be good.

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,544 @@
...
+      }
+    }
+
+    .tailwindColorUtilities(@colorName, @colorValue) {
+      // Spaces used with ^= here to prevent matching classes with dark: prefix
+
+      // Border classes
+      [class*=" border-@{colorName}"], [class^="border-@{colorName}"] {
+        border-color: @colorValue;
+      }
```
uncenter, line 64: Forgive me if I've already asked, can we not do...

```suggestion
      // Border classes
      .border-@{colorName} {
        border-color: @colorValue;
      }
```

For all of these?

Papweer: `.border-@{colorName}` doesn't work since the classes are structured like `bg-yellow-500` and I want to match any possible option for the value. I feel like there should be a way to do it nicely in less but I can't find anything. I just updated the comment to explain it better.

Papweer:
Not sure how to optimize the svg to satisfy the lint. Tried optimizing using svgo but still doesn't work.

uncenter:
> Not sure how to optimize the svg to satisfy the lint. Tried optimizing using svgo but still doesn't work.

Did you run `deno task lint:fix`?

Papweer:
> Did you run `deno task lint:fix`?

Thanks it worked, I didn't know how to run deno lint locally so it will be useful.

uncenter:
> > Did you run `deno task lint:fix`?
> 
> Thanks it worked, I didn't know how to run deno lint locally so it will be useful.

We have it documented under https://userstyles.catppuccin.com/contributing/#development-environment but I get our documentation can be confusing. If you have any suggestions for us to make that (or any) part of the docs more clear and straightforward please let me know so we can improve! 🙏

uncenter:
> Thanks it worked, I didn't know how to run deno lint locally so it will be useful.

I'm hoping https://github.com/catppuccin/userstyles/pull/1901 will help with informing people about the autofix command!

styles/hardcover/catppuccin.user.less
```diff
@@ -129,6 +129,34 @@
         .coloredElement(@colorValue);
       }
     }
+    
+    * {
```
uncenter: We should avoid using the universal selectors for performance reasons. Use whatever specific selectors the website itself likely uses.

Papweer:
Lint is passing locally but failing on Github with no error message. Lint fix does nothing. Got rid of the @property but still failing. 
Edit: Fixed now

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,580 @@
...
+    [type="checkbox"], [type="radio"] {
+      color: @accent;
+      @svg: escape(
+          '<svg viewBox="0 0 16 16" fill="@{base} xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" r="3"/></svg>'
+        );
+      background-image: url("data:image/svg+xml,@{svg}");
+    }
+
+    // Misc
+    button[aria-label="Close toast"] {
```
WalkQuackBack: This seems a bit flaky; while Hardcover only supports one language right now, this would break if another language was used.

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,580 @@
...
+      fill: @accent;
+    }
+
+    // Sections of text that for some reason dont use tailwind
+    .prose {
+      color: @text;
+      strong {
+        color: @text;
+      }
+      a:not(.text-primary-foreground) { // :not required for 2024 recap
```
WalkQuackBack: You can lose specifity on this by removing the !important and/or using a :where() selector.

styles/hardcover/catppuccin.user.less
```diff
@@ -0,0 +1,580 @@
...
+    }
+    .dark\:bg-gray-900 > div > svg {
+      // Homepage feature box icons
+      fill: @text;
+    }
+
+    // === SVG Elements ===
+
+    // Hardcover Icon
+    svg [fill="#4F46E5"],
```
WalkQuackBack: For the SVGs here, you don't need to specify SVG as a parent element. It's slower than directly using the property selector, and the chance of a non SVG element having a fill property is low.

WalkQuackBack (changes requested):

styles/hardcover/catppuccin.user.less
Papweer: Replaced with the actual selector used by hardcover
```
:where(a):not(:where([class~="not-prose"], [class~="not-prose"] *))
```

styles/hardcover/catppuccin.user.less
Papweer: Using ```:where([data-close-button])``` instead

WalkQuackBack (commented):
Just nitpicks, but aside from that everything looks good~

Image

The loading bar here seems unthemed. It should use Accent.

PR #1679, opened 2025-03-26
fix(pinterest): loading screen logo

uncenter:
Themes the bouncing pinterest logo you can see briefly when (re)loading the tab.

### DIFF
--- a/styles/pinterest/catppuccin.user.less
+++ b/styles/pinterest/catppuccin.user.less
@@ -234,6 +234,16 @@
     /* the void (parts not loaded yet) */
     background: @base !important;
 
+    /* Loading screen logo */
+    .animatedLogo {
+      [fill="#E60023"] {
+        fill: @accent !important;
+      }
+      [fill="white"] {
+        fill: @crust !important;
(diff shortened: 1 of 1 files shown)


styles/pinterest/catppuccin.user.less
```diff
@@ -275,10 +285,6 @@
       }
     }
 
-    svg circle[fill="white"] ~ path[fill="#e60023"] {
-      fill: @accent !important;
-    }
-
```
uncenter, line 281: Wasn't quite sure what this was but it references the same fill colors as for the logo above. Seems like maybe something old, for the loading logo even? If not maybe we could combine it with the stuff for the loading screen?

isabelroses (approved):

isabelroses:
works like a charm

nuexq (approved):

styles/pinterest/catppuccin.user.less
nuexq: It's old I think it was added before they put classes to the loading logo.

PR #1689, opened 2025-04-02
feat(kagi): init

cfbender:
## 🎉 Theme for [Kagi](https://kagi.com) 🎉

Kagi is a search engine that has some AI and translation features.

preview

## 💬 Additional Comments 💬

This is my first time porting something, so hopefully I didn't make any grave errors. I tried to keep things as variables where possible and keep everything consistent. I dug through as many menus and clicked as many things as I could think of but it's likely I missed at least something. Hopefully someone who is a reviewer uses Kagi or it could be difficult to see! They have a free trial so there is that option as well.

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
  - [x] `preview.webp` - composite image of all four individual flavor screenshots (taken with the default accent color of mauve) stitched together, generated via [Catwalk](https://github.com/catppuccin/catwalk).

### DIFF
--- a/scripts/userstyles.yml
+++ b/scripts/userstyles.yml
@@ -125,6 +125,8 @@ collaborators:
     url: https://github.com/nuexq
   - &WalkQuackBack
     url: https://github.com/WalkQuackBack
+  - &cfbender
+    url: https://github.com/cfbender
 
 userstyles:
   advent-of-code:
@@ -544,6 +546,13 @@ userstyles:
     readme:
       app-link: https://jisho.org
     current-maintainers: [*Lichthagel]
+  kagi:
+    name: Kagi
+    categories: [search_engine, artificial_intelligence]
+    color: yellow
+    readme:
+      app-link: https://kagi.com
+    current-maintainers: [*cfbender]
   keybr.com:
     name: keybr.com
     categories: [productivity]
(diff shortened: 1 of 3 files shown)


uncenter:
Fwiw there is already https://github.com/catppuccin/catppuccin/issues/2642, but I'm not exactly sure how these two relate. Seems like maybe we can skip the userstyle and give Kagi CSS to use directly or something?

cfbender:
@uncenter aww dang I did a quick search here but didn't think to check the main style repo. Yeah, kagi does let you inject custom CSS, I just learned this yesterday while working on this port ironically. It may just be better than requiring stylus, as it may work via account-level config but I haven't tested it. If you're worried about the duplication we don't have to proceed with this really, but I'll probably try to keep maintaining it on my fork!

uncenter:
As we have done for other ports (mdBook, Gitea), if/when the Kagi port is added upstream, we can set up a userstyle here that imports the theme from catppuccin/kagi instead of duplicating the effort here. Thanks again for your work on this!

different-name:
@cfbender for what it's worth - your theme looks a lot nicer than the upstream PR

PR #1695, opened 2025-04-06
ci: harden

isabelroses:
changes made with the help of zizmor


.github/workflows/deno-check.yml
uncenter, line 1: Changes don't seem to be warranted given the scope of the file?

.github/workflows/deno-lock.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches: [main]
     paths: ["deno.json"]
 
+permissions:
+  contents: write
```
uncenter, line 9: What does this change and why is it necessary?

.github/workflows/issues.yml
```diff
@@ -5,6 +5,8 @@ on:
   issues:
     types: [opened, edited]
 
+permissions: {}
```
uncenter, line 8: Doesn't seem necessary?

.github/workflows/lint.yml
uncenter, line 1: Changes don't seem necessary given workflow's scope.

PR #1697, opened 2025-04-09
fix(pinterest): avoid formatter syntax errors

uncenter:
Updates some syntax in the Pinterest userstyle to avoid errors the Deno formatter was throwing.

### DIFF
--- a/styles/pinterest/catppuccin.user.less
+++ b/styles/pinterest/catppuccin.user.less
@@ -209,6 +209,8 @@
     --color-background-warning-base: @peach;
     --color-background-success-base: @green;
 
+    // Currently unsupported syntax by the Less parser the Deno formatter uses, see https://github.com/g-plane/raffia/issues/9.
+    
     // @colors: {
     //   01: @blue;
     //   02: @teal;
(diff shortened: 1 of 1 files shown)


styles/pinterest/catppuccin.user.less
```diff
@@ -690,7 +715,10 @@
           background-image: url("data:image/svg+xml,@{svg}") !important;   
         }
         div[style*='background-image: url("https://s.pinimg.com/webapp/loveStatic-31fc2a99.svg")'] {
-          @svg: escape('<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 120 120"><path fill="@{red}" d="M66.72 17.122 60 23.897l-6.72-6.775c-12.196-12.163-31.975-12.163-44.171 0-12.145 12.209-12.145 32.014 0 44.223L60 111.636l50.891-50.291c12.145-12.209 12.145-32.014 0-44.223C104.793 11.041 96.799 8 88.805 8s-15.987 3.041-22.085 9.122"/><path fill="@{darken(@red, 50%)}" d="M32.727 48.909a8.182 8.182 0 1 0 16.365 0 8.182 8.182 0 0 0-16.365 0m38.182 0a8.18 8.18 0 0 0 8.182 8.182 8.182 8.182 0 1 0 0-16.364 8.18 8.18 0 0 0-8.182 8.182M60 66.8c-5.9 0-11.4-1.3-16.3-3.7.2 8.9 7.4 16 16.3 16s16.2-7.2 16.3-16c-4.9 2.3-10.4 3.7-16.3 3.7"/></svg>');
```
uncenter, line 693: @nuexq you previously attempted to interpolate with `@{darken(@red, 50%)}` - unfortunately, Less doesn't support invoking functions in the interpolation syntax, you can only include the name of a variable. This isn't your fault though, for some reason Less doesn't error when encountering this (invalid) syntax: https://github.com/less/less.js/issues/4338. (Whereas notably our Less formatter *did* error - https://github.com/g-plane/raffia/issues/10.)

styles/pinterest/catppuccin.user.less
```diff
@@ -690,7 +715,10 @@
...
+          @darkened-red: darken(@red, 50%);
```
uncenter, line 718: I should also note that darkening an accent by 50% is quite drastic 😅 do we need to modify it so heavily?

nuexq: Yeah, the darkened-red here is just for heart's smile and eyes, I went with darker red rather than just black, but we can change it to black or lighter it more (less than 45% ends up close to heart's color).

styles/pinterest/catppuccin.user.less
nuexq: Thanks for the fix I didn't notice that at all duo to the original color was also a dark red.

nuexq (approved):

PR #1708, opened 2025-04-11
fix(crates.io): update variables names

nuexq:
## 🔧 What does this fix? 🔧

- update `--csstools-light-dark-toggle--{}` variables
	- `36` -> `66`
	- `63` -> `125`
	- `92` -> `76`
	- `91` -> `75`

- added highlight js variables for code blocks

## 🗒 Checklist 🗒

### DIFF
--- a/styles/crates.io/catppuccin.user.less
+++ b/styles/crates.io/catppuccin.user.less
@@ -127,9 +127,9 @@
       --csstools-light-dark-toggle--125: @base;
 
       /* Inline code background */
-      --csstools-light-dark-toggle--76: @mantle;
+      --csstools-light-dark-toggle--76: @mantle !important;
       /* Code block background */
-      --csstools-light-dark-toggle--75: @mantle;
+      --csstools-light-dark-toggle--75: @mantle !important;
       /* Copyable text hover */
       --csstools-light-dark-toggle--78: @base;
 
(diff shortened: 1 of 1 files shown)


styles/crates.io/catppuccin.user.less
```diff
@@ -318,6 +318,52 @@
...
+    .hljs-quote             { color: @green !important; }
+    .hljs-selector-tag      { color: @yellow !important; }
+    .hljs-selector-id       { color: @blue !important; }
+    .hljs-selector-class    { color: @teal !important; }
+    .hljs-selector-attr     { color: @mauve !important; }
+    .hljs-selector-pseudo   { color: @teal !important; }
+    .hljs-template-tag      { color: @flamingo !important; }
+    .hljs-template-variable { color: @flamingo !important; }
+    .hljs-addition          { color: @green !important; }
+    .hljs-deletion          { color: @red !important; }
```
uncenter, line 366: These should be imported as per our documentation: https://github.com/catppuccin/userstyles/blob/main/docs/guide/syntax-highlighting.md.

uncenter: As a matter of fact they already are. These are unnecessary. If they don't load for you, check your browser console for errors.

uncenter:
If any variables are broken, almost all are going to need to be changed (see the duplicate variables from the linter). You'll want to check every tab of every package, find all the places each variable is used.

MuntasirSZN:
Waiting for this to be merged..... Currently turned it off. 

@uncenter @nuexq Any updates?

uncenter:
@MuntasirSZN as this website randomly generates the CSS variables every time it builds, it is near impossible and consistently frustrating to keep and maintain it. This is currently pending as I decide how to handle this in the future. If you really want to move this along the most helpful thing would be to see if upstream crates.io source can adopt consistent color variables, for the base hex codes at least, making it easier here.

uncenter:
(For example looking at https://github.com/rust-lang/crates.io/tree/main/app/styles I wonder if whatever bundler is used for the CSS modules is making this difficult...)

MuntasirSZN:
They won't, sadly @uncenter 

rust-lang/crates.io#11302

uncenter:
In the future please don't make upstream issues in that fashion. Refer it to staff if possible, and generally our approach should not be "pls make this work for our random project" - if we can generally improve their styling/build system while also improving it in such a way that makes our theming easier, that's a win-win. We can't expect them to care about niche userstyling.

uncenter:
> this website randomly generates the CSS variables every time it builds, it is near impossible and consistently frustrating to keep and maintain it.

Closing for this reason, this userstyle no longer fixes an issues since the CSS variables are constantly regenerated.

PR #1724, opened 2025-04-23
feat(google-photos): gm3 variables

WalkQuackBack:
## 🔧 What does this fix? 🔧

Refactors Google Photos to remove most generated colour classes in exchange for variables. However, this does change the look of the site matching M3 guidelines. Before and after images are attached below for reference. 

### Images
photos google com_updates(Nest Hub Max) (9)
photos google com_updates(Nest Hub Max) (8)

photos google com_updates(Nest Hub Max) (6)
photos google com_updates(Nest Hub Max) (7)

photos google com_updates(Nest Hub Max) (5)
photos google com_updates(Nest Hub Max) (4)

photos google com_updates(Nest Hub Max) (2)
photos google com_updates(Nest Hub Max) (3)

photos google com_updates(Nest Hub Max) (1)
photos google com_updates(Nest Hub Max)

I would also like to add myself as a maintainer of the userstyle.

## 🗒 Checklist 🗒


styles/google-photos/catppuccin.user.less
```diff
@@ -67,867 +84,608 @@
...
-    .rniux {
-      color: @subtext0;
-    }
-
-    /* slideshow buttons */
-    .UPRhTe {
-      background-color: @base;
-    }
-
-    /* activity log page */
```
WalkQuackBack, line 688: I wasn’t sure if it was worthwhile to theme, considering Google themeselves don’t even theme it when dark theme is enabled.

styles/google-photos/catppuccin.user.less
```diff
@@ -15,16 +15,31 @@
...
       #catppuccin(@lightFlavor);
     }
-    @media (prefers-color-scheme: dark) {
+  }
+
+  @media (prefers-color-scheme: dark) {
+    :root {
       #catppuccin(@darkFlavor);
     }
   }
```
uncenter, line 33: What's the reason for the change?

styles/google-photos/catppuccin.user.less
```diff
@@ -15,16 +15,31 @@
...
 
+  .zKHdkd {
+    #catppuccin(@lightFlavor);
+  }
+
+  .dm7YTc {
+    #catppuccin(@darkFlavor);
+  }
```
uncenter, line 41: Can we avoid autogenerated classnames?

styles/google-photos/catppuccin.user.less
```diff
@@ -54,7 +69,9 @@
     @crust: @catppuccin[@@flavor][@crust];
     @accent: @catppuccin[@@flavor][@@accentColor];
 
-    color-scheme: if(@flavor = latte, light, dark);
+    @inverseTheme: if(@flavor = latte, mocha, latte);
+    @inverseAccent: @catppuccin[@@inverseTheme][@@accentColor];
```
uncenter, line 73: Let's use kebab-case for Less variables please (helps distinguish from the metadata vars which are camelCase).
```suggestion
    @inverse-theme: if(@flavor = latte, mocha, latte);
    @inverse-accent: @catppuccin[@@inverse-theme][@@accentColor];
```

styles/google-photos/catppuccin.user.less
```diff
@@ -67,867 +84,608 @@
...
+      @surface2,
+      @surface1,
+      @surface0,
+      @base,
+      @mantle,
+      @crust,
+      @accent,
+      @inverseTheme,
+      @inverseAccent,
+    ) {
```
uncenter, line 117: A bit different than we usually do but I'd be okay if you want to just pass in the palette (e.g. `@catppuccin[@@flavor]` or `@mocha` like you have set below) into this mixin, and move (copy? are they used anywhere else) the variable declarations for each of the colors above into this function - if that works. Lmk if that makes sense lol.

styles/google-photos/catppuccin.user.less
```diff
@@ -15,16 +15,31 @@
 @var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
 ==/UserStyle== */
 
+#rgbify(@color) {
+  @rgb: red(@color), green(@color), blue(@color);
+}
```
uncenter, line 20: Sorry for the nitpick - can you move this to the end of the `@-moz-document` and above the `@catppuccin` palette? I like to keep the top of the userstyle simple and leave generic mixins/"builtins" at the bottom.

uncenter (commented):
This is pretty incredible that we can use variables for all this, great work.

styles/google-photos/catppuccin.user.less
WalkQuackBack: Not avoidable unfortunately, that's the only selector of whether dark mode is enabled or not in Google Photos settings. Before the refactor, the theme setting was ignored.
image
image

styles/google-photos/catppuccin.user.less
WalkQuackBack: Mostly understand, will push and you could verify that it's what you want!

styles/google-photos/catppuccin.user.less
WalkQuackBack: Not sure, actually. Can revert it.

styles/google-photos/catppuccin.user.less
```diff
@@ -560,105 +562,8 @@
...
-        @mocha[@surface1],
-        @mocha[@surface0],
-        @mocha[@base],
-        @mocha[@mantle],
-        @mocha[@crust],
-        @mocha[@@accentColor],
-        latte,
-        @catppuccin[@latte][@@accentColor],
-      );
-    }
```
uncenter: Did you intend to remove this fully? Unless I'm mistaken I don't see code producing the same output added on the rright 🤔

styles/google-photos/catppuccin.user.less
uncenter: Yep you got what I meant exactly, thanks!

styles/google-photos/catppuccin.user.less
WalkQuackBack: Yes, it was mistakenly duplicated.

styles/google-photos/catppuccin.user.less
uncenter: Do you mind making this revert change?

styles/google-photos/catppuccin.user.less
uncenter: Well now the only `@theme` you are passing in to the mixins is `@catppuccin[@@flavor]`/the current flavor palette, correct me if I'm wrong? And you only use each mixin once? So why not drop the mixin and the extra variable declarations?

styles/google-photos/catppuccin.user.less
WalkQuackBack: Will do.

styles/google-photos/catppuccin.user.less
WalkQuackBack: Will inline the code then: but need to do another pass to double check the code wasn’t necessary at all.

styles/google-photos/catppuccin.user.less
```diff
@@ -17,14 +17,23 @@
...
-    @media (prefers-color-scheme: dark) {
+    #catppuccin(@lightFlavor);
+  }
+
+  @media (prefers-color-scheme: dark) {
+    :root {
       #catppuccin(@darkFlavor);
     }
   }
```
uncenter, line 27: I still don't see why we can't just stick with the original code on the left here.

WalkQuackBack: I didn’t realize my revert from the original implementation, but it should do the same thing from my testing

uncenter: It doesn't do the same thing, this always adds the light flavored code to the injected styles regardless of what the preferred color scheme is. It is unnecessary. Please revert as I've asked before.

WalkQuackBack: Is there a good way to revert certain sections of code to a previous commit aside from copy pasting from the old hash?

uncenter: You are severely overthinking this... did it for you.

CleanShot 2025-04-30 at 17 16 49

WalkQuackBack: Thanks.

GenShibe (approved):
LGTM, thanks for doing all this lol

styles/google-photos/catppuccin.user.less
GenShibe: then i think we can remove the theming for it

uncenter (approved):
I don't use Google Photos at all but from what I can see on my end everything looks themed great. Nicely done.

styles/google-photos/catppuccin.user.less
```diff
@@ -54,7 +62,9 @@
     @crust: @catppuccin[@@flavor][@crust];
     @accent: @catppuccin[@@flavor][@@accentColor];
 
-    color-scheme: if(@flavor = latte, light, dark);
+    @inverse-theme: if(@flavor = latte, mocha, latte);
+    @inverse-accent: @catppuccin[@@inverse-theme][@@accentColor];
+    @color-scheme: if(@flavor = latte, light, dark);
```
uncenter, line 67: I don't think we meant to change this to a Less variable right? Looking back on it in https://github.com/catppuccin/userstyles/pull/1846/files#diff-2011c215f12927338a258d8a62d940c81135e8ca5f8419516925e5a56935f925...

WalkQuackBack: Definitely looks unintentional, but since Google Photos uses entirely custom components I'm not sure it impacts anything visually (albeit, am unsure about the scrollbars)

uncenter: I was going to ask that you test if it has a difference but I just went to https://photos.google.com and it looks like most of the UI is unthemed right now. Can you replicate that on your end?

WalkQuackBack: Looks like the selector change from :root to body, I'll fix that quickly in a new PR.

uncenter: Now that that is fixed, can you test if anything breaks by setting `color-scheme` properly with the defaults?

PR #1726, opened 2025-04-24
feat(wikipedia): support various other public mediawikis

uncenter:
re: https://github.com/catppuccin/catppuccin/discussions/2771 @scarcekoi


gradylink:
Would this work with https://osdev.org or https://osdev.wiki?

uncenter:
> Would this work with [wiki.osdev.org](https://wiki.osdev.org) or [osdev.wiki](https://osdev.wiki)?

Non Wikimedia ones should be added via https://github.com/catppuccin/userstyles/blob/main/docs/USAGE.md#decentralizedself-hosted-applications.

uncenter:
Any other unthemed areas are likely to be from the website's failure to properly use Wikipedia color variables and I'd like to avoid custom CSS to fix that. This fixes some basic problems and themes the logos/titles in the top left.

Issue #1546, opened 2024-12-30
Recent Major/Breaking Changes


uncenter:
Hey all! Happy holidays! The staff team has been making a few major and/or breaking changes in the last few days, this issue serves as a summary of what users and maintainers each need to know.

## User-facing Changes

We have recently made a change in #1544 that renamed all userstyle source files, which is not handled well by Stylus. To remedy this, we have created "stub" userstyles that will serve as an intermediary between the old filename and the new filename. If you update your userstyles in Stylus, or if they get auto-updated in the background, they will no longer work and will say [STUB | UPDATE REQUIRED] in the title. As indicated by the updated title, a second, additional update is required to restore the userstyles to a working state. You can bulk update them all at once by clicking the following button on the Stylus manage page:

![CleanShot 2024-12-30 at 14 26 27](https://github.com/user-attachments/assets/c8be40b4-7b5c-44e6-aa13-26a26f8a8789)

Alternatively, it might be quicker to reinstall all userstyles using our generated Stylus import/export file: https://github.com/catppuccin/userstyles/blob/main/docs/USAGE.md#all-userstyles.

Our apologies for the breaking change, and thank you for your understanding!

## Developer/Maintainer-facing Changes

As part of #1544, the Prettier formatter was completely removed from our tooling for this repository. Userstyles should now by formatted by Deno's built in formatter, `deno fmt`, as are all other files. Additionally, because of the source file renames, existing pull requests may need to be updated to edit the correct file and init pull requests will need to rename the new userstyle file and update metadata.

A smaller, quality-of-life change was recently merged as part of #1534/#1535 to switch the userstyle versioning system from "SemVer" to "CalVer". Contributors to this repository are no longer required to and should not update the version of modified userstyles in pull requests - this is now done automatically by GitHub Actions.

uncenter:
I should note that we do not plan on making any other breaking changes! The mass rename was a one-time-thing and I can't imagine we will ever need to perform a similarly breaking change again.

That being said, we do have one last major - not breaking - change before the end of the year, for developers/maintainers in #1547. Aims to clean up numerous accumulated bugs, and simplify other areas of development. The `@accent-color` variable has been simplified to `@accent`, `@lookup` to `@flavor`, duplicate media queries were simplified, the recently documented `@catppuccin-filters` theming method for images was implemented, and some readability lints were added. **Users should see no visible changes to userstyle-themed websites.**

PR #1732, opened 2025-04-29
fix(bsky): highlighted notifications

teidesu:
as of recently bsky started using `<a>` for notifications, which broke the currently used selectors
image

`<div>` is still used for *some* notifications (e.g. replies), so i kept that as well.


styles/bsky/catppuccin.user.less
```diff
@@ -2,7 +2,7 @@
 @name Bluesky Social Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/bsky
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/bsky
-@version 2024.12.31
+@version 2025.04.29
```
uncenter, line 5: No need to update the version, this is updated automatically by GitHub Actions after merging, see #1546.

RoootTheFox (approved):
lgtm, works and is fine besides the version thing uncenter mentioned c:

<sub>(also, quick note to userstyles maintainers: once that issue is sorted, feel free to merge this right away as i likely won't be there at all until the end of next week, as i'll be busy with my school finals ^^)</sub>

styles/bsky/catppuccin.user.less
```diff
@@ -538,6 +538,9 @@
     div[style*="background-color: rgb(0, 17, 36)"], // dark
     div[style*="background-color: rgb(21, 41, 63)"], // dim
     div[style*="background-color: rgb(240, 247, 255)"] // light
+    a[style*="background-color: rgb(0, 17, 36)"], // dark
+    a[style*="background-color: rgb(21, 41, 63)"], // dim
+    a[style*="background-color: rgb(240, 247, 255)"] // light
```
uncenter, line 543:
```suggestion
   [style*="background-color: rgb(0, 17, 36)"], // dark
   [style*="background-color: rgb(21, 41, 63)"], // dim
   [style*="background-color: rgb(240, 247, 255)"] // light
```

How about just simplifying the selector like this?

styles/bsky/catppuccin.user.less
uncenter, line 5:
```suggestion
@version 2024.12.31
```

styles/bsky/catppuccin.user.less
RoootTheFox: would work, though i feel like keeping a wildcard selector specific to 2 types of elements (instead of *all*) is a better idea performance-wise (obviously only if done a lot + i haven't actually benchmarked that), im fine with either way though
(reasoning: i wanna avoid my styles becoming similar to the discord style in terms of performance)

styles/bsky/catppuccin.user.less
```diff
@@ -538,6 +538,9 @@
     div[style*="background-color: rgb(0, 17, 36)"], // dark
     div[style*="background-color: rgb(21, 41, 63)"], // dim
     div[style*="background-color: rgb(240, 247, 255)"] // light
```
RoootTheFox, line 540: on a different note, there's a missing comma here which might mess things up considering all of these should be a list selector ^^

RoootTheFox (changes requested):

styles/bsky/catppuccin.user.less
teidesu: oops lol

RoootTheFox (approved):

PR #1741, opened 2025-05-03
fix(youtube): distinguish music repeat/shuffle off from on

pynappo:
## 🔧 What does this fix? 🔧

Makes it much easier to tell whether shuffle/repeat is off vs on when using youtube music

before:

on and off look identical:
image

after:

this is off:

image

on looks like before:

image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/youtube/catppuccin.user.less
+++ b/styles/youtube/catppuccin.user.less
@@ -1541,6 +1541,51 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


isabelroses (commented):
This should be submitted to https://github.com/catppuccin/youtubemusic instead

pynappo:
Oh i didn't know that existed tbh, nvm this then

PR #1744, opened 2025-05-06
fix(canvas-lms): tabs, toggles, equation editor, misc.

pynappo:
## 🔧 What does this fix? 🔧

- style tabs/toggle switches better
- support for the built-in equation editor
- some button fixes (e.g. on tinymce, the use webcam button on file upload, and the buttons on mobile nav menu)

## 🗒 Checklist 🗒

### DIFF
--- a/styles/canvas-lms/catppuccin.user.less
+++ b/styles/canvas-lms/catppuccin.user.less
@@ -647,38 +647,44 @@
     }
 
     /* checkboxes */
-    input + label {
-      [class*="checkboxFacade__facade"],
-      [class*="toggleFacade__facade"] {
-        border-color: @surface1;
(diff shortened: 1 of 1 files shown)


styles/canvas-lms/catppuccin.user.less
```diff
@@ -347,6 +342,13 @@
       }
     }
 
+    div[aria-label="Global Navigation"]:not([class]) {
```
uncenter, line 345: `:not([class])`?

pynappo: the mobile navigation menu doesn't have a class while the desktop one does but i don't think this extra attribute matters anyways

uncenter (approved):

PR #1747, opened 2025-05-08
fix(gmail): tab selector

yaner-here:
## 🔧 What does this fix? 🔧

Gmail frontend has been updated. It disabled the tab selector with a new confused `classname`: `.Tm > .ya` → `.Tm > .aeF`. It's better to select the parent element `.Tm` directly.

It's promised that there's only one `.Tm` element.
image

### Before:

image

### After:

image

## 🗒 Checklist 🗒

- [-] I have read and followed Catppuccin's [contributing guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md).

### DIFF
--- a/styles/gmail/catppuccin.user.less
+++ b/styles/gmail/catppuccin.user.less
@@ -581,7 +581,7 @@
     }
     /* No drafts message, trash 30 days message */
     .TD,
-    .Tm .ya {
+    .Tm {
       background-color: @surface1;
       color: @text;
     }


uncenter:
I don't see that UI change on my end yet, and that selector isn't meant for tabs, I think it's meant for the "No drafts" and "trash 30 days" messages as suggested by the comment above.

yaner-here:
> I don't see that UI change on my end yet

Is your output the same as mine? Now the frontend DOM tree has no matching elements with the original selector `.Tm .ya`.

```javascript
/* Browser Devtools Console */

>  document.querySelectorAll(".Tm .ya")
<- NodeList []

>  document.querySelectorAll(".Tm")[0].childNodes
<- NodeList(2) [div#:1.aeF, div#:2.aeG]
```

`.Tm > .aef` represents the top tab and entry container, `.Tm > .aeG` represents the footbar.

> that selector isn't meant for tabs, I think it's meant for the "No drafts" and "trash 30 days" messages as suggested by the comment above.

I think the comments are outdated and misleading. Also, there is no matching elements with the original selector `.TD`. Perhaps the classname has a new reference object after compilation and obfuscation, at least it's definitely not so called `No drafts message` and `trash 30 days message`.

```javascript
/* Browser Devtools Console */

>  document.querySelectorAll(".TD")
<- NodeList []
```

Please inform me if my reply convinces you. I'm glad to update the comments.

uncenter:
> Also, there is no matching elements with the original selector `.TD`. Perhaps the classname has a new reference object after compilation and obfuscation, at least it's definitely not so called `No drafts message` and `trash 30 days message`.

Are you on a page where the "no drafts" message is present? 😅

yaner-here:
Oh sorry, I see `.TD`. Since it works in your browser, just leave it.

PR #1749, opened 2025-05-09
chore(keybr.com): change category to `education`

sgoudham:
I was adding [catppuccin/keybr](https://github.com/catppuccin/keybr) and I think `education` makes more sense.

### DIFF
--- a/scripts/userstyles.yml
+++ b/scripts/userstyles.yml
@@ -442,7 +442,7 @@ userstyles:
   keybr.com:
     name: keybr.com
     link: https://keybr.com
-    categories: [productivity]
+    categories: [education]
     color: text
     current-maintainers: [*TadoTheMiner]
   keyoxide:


scripts/userstyles.yml
```diff
@@ -442,7 +442,7 @@ userstyles:
   keybr.com:
     name: keybr.com
     link: https://keybr.com
-    categories: [productivity]
+    categories: [education]
```
sgoudham, line 445:
```suggestion
    categories: [productivity, education]
```

uncenter:
I'm still not convinced this lines up with our other uses of the category so I'm inclined to close this...

sgoudham:
I wasn't really expecting it to invoke this amount of discussion lol. I think it's a valid categorisation and it would match up with what I did over in https://github.com/catppuccin/catppuccin/blob/91b3f8d9f8cd2f0c30247b2a89df252d45723479/resources/ports.yml#L1062-L1064.

But I'm also not going to die on this hill lol, you can close it if you're really against it.

PR #1787, opened 2025-06-12
fix(chatgpt): main bg

ashish0kumar:
## 🔧 What does this fix? 🔧

Fixes unthemed main background

|Before|After|
|---|---|
|image|image|

## 🗒 Checklist 🗒

### DIFF
--- a/styles/chatgpt/catppuccin.user.less
+++ b/styles/chatgpt/catppuccin.user.less
@@ -122,6 +122,10 @@
       background-color: @mantle;
     }
 
+    #main {
+      background-color: @base !important;   
+    }
+
     color: @text;
 
     @white: if(@flavor = latte, @crust, @text);


uncenter (changes requested):
I'd appreciate if we could take a more in-depth look at these kinds of fixes. Your solution might work, but it isn't the *right* way of doing it. If you look closely, it appears that we only need to set the `--bg-primary` variable to fix the background. (When a userstyle accumulates lots of haphazard fixes like this one, confusion around where to change things and what code is still needed sets in!)

ashish0kumar:
> I'd appreciate if we could take a more in-depth look at these kinds of fixes. Your solution might work, but it isn't the _right_ way of doing it. If you look closely, it appears that we only need to set the `--bg-primary` variable to fix the background. (When a userstyle accumulates lots of haphazard fixes like this one, confusion around where to change things and what code is still needed sets in!)

You're right, apologies! I’ll keep this in mind and make sure to go for more precise fixes moving forward. Thanks for pointing it out!

uncenter (approved):

Issue #1411, opened 2024-10-29
Whatsapp Web: Pin text from green to purple


mambucodev:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:whatsapp-web

### Describe your problem.

I'd suggest to color that "Pinned"  banner with catppuccin instead of green, it would be much smoother.

Mantainer: @FrancoRav 

### Attach screenshots.

![image](https://github.com/user-attachments/assets/ca5132d0-6026-4d31-a71c-36836f866f4a)

### What browser(s) are you seeing the problem on?

Firefox v131.0.3

TimoNotThy:
Some more like this:
* In the status tab the word **RECENT** and the **end-to-end-encrypted** statement
* In the channels tab the word **follow** inside of the buttons
* At the beginning of a chat the **safety tools** text and the **Learn more** button when clicking safety tools
* The text about messages being end-to-end encrypted at the start of a chat

uncenter:
Hi, sorry for the delay on this. @mambucodev @TimoNotThy are these issues still present if you update now? We recently merged https://github.com/catppuccin/userstyles/pull/1798 and I think it themes everything completely.

mambucodev:
> Hi, sorry for the delay on this. @mambucodev @TimoNotThy are these issues still present if you update now? We recently merged https://github.com/catppuccin/userstyles/pull/1798 and I think it themes everything completely.

I'm sorry but I'm not using the theme anymore, nor any other userstyle. of nobody else has this issue you can close it.

uncenter:
No worries, thanks for letting me know!

Issue #881, opened 2024-05-10
Color issues


bernzrdo:
### Is there an existing issue outlining your problem?

- [X] I am using the latest version of the provided userstyle.
- [X] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:whatsapp-web

### Describe your problem.

I've collected some places where wrong colors are being used and their associated variables.

### `--button-secondary`
"View all chats" button when there are no unread.
![image](https://github.com/catppuccin/userstyles/assets/27358071/7c7742d7-553a-4b50-b86d-46579b37c675)
Block dialog (and possibly other dialogs)
![image](https://github.com/catppuccin/userstyles/assets/27358071/e7f07f08-8817-401a-8055-cf54d6102e05)

### `--icon-primary`
Pinned messages banner on a chat.
![image](https://github.com/catppuccin/userstyles/assets/27358071/b3b5f901-9bd9-4169-812e-a720deec9166)

### `--teal-lighter`
No communities screen. "See example communities" uses this variable. (SVG could be modified too maybe)
![image](https://github.com/catppuccin/userstyles/assets/27358071/d308e0fe-3ed3-47e8-96fa-3122c8f75543)

### `--background-lightest-hover`
Settings hover. Mentioned on #871.
![image](https://github.com/catppuccin/userstyles/assets/27358071/1e0b9e29-6657-4cb6-8cea-0a0cfe9c3dec)

### `--teal`
Selected emojis, GIFs, stickers tab.
![image](https://github.com/catppuccin/userstyles/assets/27358071/997c6805-2382-4a25-958f-5d9e189c3ed8)
New chat screen, contact list
![image](https://github.com/catppuccin/userstyles/assets/27358071/8b27e0a9-3a17-49a0-82fe-53655a18d5dc)
Disappearing messages settings. `--text-teal` is affecting the "Make messages in this chat disappear" text.
![image](https://github.com/catppuccin/userstyles/assets/27358071/8c9656f6-7db6-4ce7-bd11-de75aa662e69)
Profile settings
![image](https://github.com/catppuccin/userstyles/assets/27358071/4198cc63-7c0b-49f6-b69d-ac1ee3e31bc0)
Settings > Privacy
![image](https://github.com/catppuccin/userstyles/assets/27358071/09929a31-c9a0-4723-9557-1616b4f78576)

### `--input-empty-value-placeholder`
Empty group description
[...]

PR #1798, opened 2025-07-02
fix(whatsapp-web): rewrite theme with variables

### LINKED ISSUE #871: Settings menu un-styled menu-item hover color

- [X] I am using the latest version of the provided userstyle.

lbl:whatsapp-web

### Describe your problem.

In the settings menu, hovering over menu-items displays un-styled color.

The css variable responsible is: `--background-lightest-hover`

### Attach screenshots.

image

Firefox v125.0.2

### Any additional comments?

_No response_

### LINKED ISSUE #1754: WhatsApp has uncolored elements

lbl:whatsapp-web

### Describe your problem.

Elements in the sidebar are unthemed in lots of tabs

### Attach screenshots.

Chats tab:
Image
Settings tab:
Image

Firefox v138.0.1

### Any additional comments?

_No response_

WalkQuackBack:
## 🔧 What does this fix? 🔧

Fixes WhatsApp web colours after redesign. Some thing are missing, but not necessarily regressed and will be resolved in a follow up PR.

Closes #1754
Closes #871

## 🗒 Checklist 🗒


styles/whatsapp-web/catppuccin.user.less
```diff
@@ -67,613 +79,277 @@
...
-      > div._3gYev
-      > div
-      > div
-      > button
-      > div._3xdht._1ZD3q
-      > span
-      > svg
-      > path {
+      @WDS-systems-bubble-surface-incoming: @surface1;
+    }
```
uncenter: You can make this an inline if statement for readability.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -67,613 +79,277 @@
...
+    --focus-animation-deeper: @focus-animation-deeper;
+    --focus-animation-deeper-rgb: #rgbify(@focus-animation-deeper)[];
+    
+    /* "About AI messages" icon */
+    .st0 {
       fill: @accent !important;
+      stroke: @surface0;
+    }
+    /* Toast action (Normally yellow) */
+    .x17daq4n {
```
uncenter: Is there no other selector that isn't autogenerated?

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -67,613 +79,277 @@
...
+    input[type="date"]::-webkit-datetime-edit-day-field:focus {
+      background-color: @WDS-accent;
+      color: @WDS-content-on-accent;
     }
   }
 }
 
 #rgbify(@color) {
-  @rgb: red(@color) green(@color) blue(@color);
+  @rgb: red(@color), green(@color), blue(@color);
```
uncenter, line 299: Please add a note if/because the commas are necessary - this is a "builtin" we copy and paste around and if it is not mentioend in a comment we might change it without noticing.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -682,4 +358,4 @@
...
   @macchiato: { @rosewater: #f4dbd6; @flamingo: #f0c6c6; @pink: #f5bde6; @mauve: #c6a0f6; @red: #ed8796; @maroon: #ee99a0; @peach: #f5a97f; @yellow: #eed49f; @green: #a6da95; @teal: #8bd5ca; @sky: #91d7e3; @sapphire: #7dc4e4; @blue: #8aadf4; @lavender: #b7bdf8; @text: #cad3f5; @subtext1: #b8c0e0; @subtext0: #a5adcb; @overlay2: #939ab7; @overlay1: #8087a2; @overlay0: #6e738d; @surface2: #5b6078; @surface1: #494d64; @surface0: #363a4f; @base: #24273a; @mantle: #1e2030; @crust: #181926; };
   @mocha:     { @rosewater: #f5e0dc; @flamingo: #f2cdcd; @pink: #f5c2e7; @mauve: #cba6f7; @red: #f38ba8; @maroon: #eba0ac; @peach: #fab387; @yellow: #f9e2af; @green: #a6e3a1; @teal: #94e2d5; @sky: #89dceb; @sapphire: #74c7ec; @blue: #89b4fa; @lavender: #b4befe; @text: #cdd6f4; @subtext1: #bac2de; @subtext0: #a6adc8; @overlay2: #9399b2; @overlay1: #7f849c; @overlay0: #6c7086; @surface2: #585b70; @surface1: #45475a; @surface0: #313244; @base: #1e1e2e; @mantle: #181825; @crust: #11111b; };
-};
+};
```
uncenter: Please add the file's trailing newline back. You can do this by setting up your editor with EditorConfig and our formatter, Deno.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -67,613 +79,277 @@
...
+    @WDS-components-surface-nav-bar: @base;
+    
+    @WDS-app-wash: @surface0;
+    
+    --WDS-accent: @WDS-accent;
+    --WDS-accent-RGB: #rgbify(@WDS-accent)[];
+    --WDS-accent-deemphasized: @WDS-accent-deemphasized;
+    --WDS-accent-deemphasized-RGB: #rgbify(@WDS-accent-deemphasized)[];
+    --WDS-accent-emphasized: @WDS-accent-emphasized;
+    --WDS-accent-emphasized-RGB: #rgbify(@WDS-accent-emphasized)[];
```
uncenter: I'd rather we create a mixin for this than have basically three lines per color (one to define the Less variable, one for the simple CSS variable, one for the RGB CSS variable). You could do:

```less
#WDS(@name, @color) {
    --WDS-@{name}: @color;
    --WDS-@{name}-RGB: #rgbify(@color)[];
}
```

^ ~~Haven't tested~~ Tested basically [in the playground](https://lesscss.org/less-preview/#eyJjb2RlIjoiI2xpYigpIHtcbiAgICAuY29sb3JzKCkge1xuICAgICAgQHByaW1hcnk6IGJsdWU7XG4gICAgICBAc2Vjb25kYXJ5OiBncmVlbjtcbiAgICB9XG4gICAgLnJ1bGVzKEBzaXplKSB7XG4gICAgICBib3JkZXI6IEBzaXplIHNvbGlkIHdoaXRlO1xuICAgIH1cbiAgfVxuICBcbiAgLmJveCB3aGVuICgjbGliLmNvbG9yc1tAcHJpbWFyeV0gPSBibHVlKSB7XG4gICAgd2lkdGg6IDEwMHB4O1xuICAgIGhlaWdodDogKCR3aWR0aCAvIDIpO1xuICB9XG4gIFxuICAuYmFyOmV4dGVuZCguYm94KSB7XG4gICAgQG1lZGlhIChtaW4td2lkdGg6IDYwMHB4KSB7XG4gICAgICB3aWR0aDogMjAwcHg7XG4gICAgICAjbGliLnJ1bGVzKDFweCk7XG4gICAgfVxuICB9IiwiYWN0aXZlVmVyc2lvbiI6IjQuMy4wIiwibWF0aCI6InBhcmVucy1kaXZpc2lvbiIsInN0cmljdFVuaXRzIjpmYWxzZX0=) but off the top of my head that should work.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -67,613 +79,277 @@
...
+    @butterbar-blue-nux-background: @sky;
+    @butterbar-blue-nux-icon-background: darken(@sky, 30%);
+    @blue-light: @blue;
+
+    @gray-500: @WDS-content-deemphasized;
+
+    @focus-animation-deeper: mix(@accent, @base, 30%);
+    
+    --attachment-type-stickers-color: @attachment-type-stickers-color;
+    --attachment-type-stickers-color-rgb: #rgbify(@attachment-type-stickers-color)[];
```
uncenter: Should do a mixin down here too, not sure what to call it though.

styles/whatsapp-web/catppuccin.user.less
WalkQuackBack: Don't know exactly what you're referring to, but for the toast action, it can't really be identified through other means than the clas name. It's not essential though,.so it could be removed if that would be better.

styles/whatsapp-web/catppuccin.user.less
WalkQuackBack: Will add a comment about the changed behaviour.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -80,258 +82,114 @@
...
     
     /* "About AI messages" icon */
     .st0 {
       fill: @accent !important;
       stroke: @surface0;
     }
-    /* Toast action (Normally yellow) */
-    .x17daq4n {
-      color: @accent;
-    }
```
uncenter: If this actually themes something, then by all means keep it. I would just much rather look for a different selector, even if its something like `body > div:last-of-type()`.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -66,6 +66,8 @@
     @mantle: @catppuccin[@@flavor][@mantle];
     @crust: @catppuccin[@@flavor][@crust];
     @accent: @catppuccin[@@flavor][@@accentColor];
+    
+    @accentText: if(@flavor = latte, #FFF, darken(saturate(@accent, 100%), 65%));
```
uncenter: Can you explain? I don't understand why this is necessary.

WalkQuackBack: Trying to hit WCAG AAA for the contrast while not looking too out of place

styles/whatsapp-web/catppuccin.user.less
WalkQuackBack: Couldn't find a great way to consistently latch onto it using other selectors. You trigger the toast when performing an action like Mark as Read/Unread, but it seems to be able to be inserted at different places in the DOM presumably based off of context. Maybe you could take a look?

WalkQuackBack:
Not sure why the metadata check is failing now. Don't believe anything changed there?

styles/whatsapp-web/catppuccin.user.less
uncenter: Why is this necessary? When/where is this used with what background color? It is not your job to modify colors for WCAG, that's up to the palette and we shouldn't patch it in every userstyle separately (shades might help with this?). Anyways I need to know the context, I don't think we want this.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -67,613 +81,144 @@
...
 
+#WDS(@name, @color) {
+    --WDS-@{name}: @color;
+    --WDS-@{name}-RGB: #rgbify(@color)[];
+}
+
+#RGB(@name, @color) {
+    --@{name}: @color;
+    --@{name}-rgb: #rgbify(@color)[];
+}
```
uncenter:
```suggestion
#WDS(@id, @color) {
    --WDS-@{id}: @color;
    --WDS-@{id}-RGB: #rgbify(@color)[];
}

#RGB(@id, @color) {
    --@{id}: @color;
    --@{id}-rgb: #rgbify(@color)[];
}
```

This should fix the metadata error. I think it's an issue with the UserCSS metadata parser we use or something where because we use `@name` down here it gets confused.

styles/whatsapp-web/catppuccin.user.less
WalkQuackBack: > Why is this necessary? When/where is this used with what background color? It is not your job to modify colors for WCAG, that's up to the palette and we shouldn't patch it in every userstyle separately (shades might help with this?). Anyways I need to know the context, I don't think we want this.

It's an inverse colour, dark in light theme and light in dark. When I assigned it to @crust there was contrast issues on Latte in some of the places it was used, and before I changed it to manipulate the accent colour it was a switch between #FFF and #000

WalkQuackBack: Would it be best to make it switch to a more contrasty colour in the palette for each theme instead?

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -55,6 +107,8 @@
     @crust: @catppuccin[@@flavor][@crust];
     @accent: @catppuccin[@@flavor][@@accentColor];
 
+    @accentContent: if(@flavor = latte, @base, @crust);
```
uncenter, line 110: I still question if this is necessary? We don't do it in any other userstyles - if we want to really differentiate between using base and crust for contrast on accent backgrounds we should standardize on something through a separate process like an issue/RFC type discussion.

uncenter (changes requested):

styles/whatsapp-web/catppuccin.user.less
WalkQuackBack: WhatsApp uses the M3 design, but the on accent colour varies from userstyle to userstyle. As an example, `google-gemini` userstyle uses `@darken(@accent, 60%)` ,  `google-drive` uses `base`. I agree that the on accent/pastel colour design should be standardized to reduce inconsistencies in that regard across the whole organization.

styles/whatsapp-web/catppuccin.user.less
```diff
@@ -68,612 +122,169 @@
...
-      > g
-      > path:nth-child(2) {
-      fill: @sky !important;
+    #WDS(accent, @accent);
+    #WDS(accent-deemphasized, fade(@accent, @deemphasize));
+    #WDS(accent-emphasized, lighten(@accent, @emphasize));
+
+    #WDS(secondary-negative, @red);
+    #WDS(secondary-negative-deemphasized, fade(@red, @deemphasize));
+    #WDS(secondary-negative-emphasized, lighten(@red, @emphasize));
```
uncenter:
```suggestion
    #WDS(secondary-negative-emphasized, @red);
```

The text here looks much better to me if we just use normal red instead of twisting it.

| Before | After |
| --- | --- |
| CleanShot 2025-07-16 at 07 39 00 | CleanShot 2025-07-16 at 07 38 51 |

WalkQuackBack:
> New chat icons? Seems to be themed by `color: var(--white)` which seems fixable.
> 
> CleanShot 2025-07-16 at 07 46 10 

If I recall correctly the white colour is also used in other places, but I could probably change it and do a pass to double check that nothing else breaks. I personally don't think theming those is too much of a concern considering the stock WhatsApp dark theme doesn't touch it and it results in bad contrast.

uncenter:
> If I recall correctly the white colour is also used in other places, but I could probably change it and do a pass to double check that nothing else breaks. I personally don't think theming those is too much of a concern considering the stock WhatsApp dark theme doesn't touch it and it results in bad contrast.

Should we change this? Trying to remember what else is blocking this.

WalkQuackBack:
> > If I recall correctly the white colour is also used in other places, but I could probably change it and do a pass to double check that nothing else breaks. I personally don't think theming those is too much of a concern considering the stock WhatsApp dark theme doesn't touch it and it results in bad contrast.
> 
> Should we change this? Trying to remember what else is blocking this.

I sort of messed up the Git history and some of my fixes ended up on #1808, and can't move/recommit them over right now. I should be able to in about two weeks when I'm less busy.

WalkQuackBack:
Don't remember what the variable for white is named, and can't check it which is why it hasn't been added yet. Can you review as is for now?

uncenter:
> Don't remember what the variable for white is named

Isn't it just...?

> Seems to be themed by `color: var(--white)` which seems fixable.

uncenter:
Adding `#WDS(white, @crust);` fixes the new chat icons and doesn't seem to break anything else. Do you want to add it and test it out and if that works we can merge this?

uncenter (approved):

uncenter:
Since @FrancoRav hasn't been involved yet and these changes are solid, I'm good to merge this. @WalkQuackBack are you interested in becoming a maintainer for this userstyle as well?

uncenter:
I linked https://github.com/catppuccin/userstyles/issues/871 as closed by this pull request since it seems to not have those issues with these changes. There are also https://github.com/catppuccin/userstyles/issues/1411 and https://github.com/catppuccin/userstyles/issues/881 which I will comment on after this merges to see if the authors still have any issues.

uncenter:
> I'm good to merge this. @WalkQuackBack are you interested in becoming a maintainer for this userstyle as well?

I'll merge now, feel free to raise a pull request later adding yourself as a maintainer if you are willing!

alba4k:
@WalkQuackBack not sure what happened upstream but the issue this fixed came back, I think?
immagine

WalkQuackBack:
> @WalkQuackBack not sure what happened upstream but the issue this fixed came back, I think? immagine

There was a hash reroll, which I forgot to account for in the initial userstyle. I'll see if there's another way to select the root theme element that's resistant to those.

styles/whatsapp-web/catppuccin.user.less
uncenter: Okay, silly me here - the builtin version that we use (now in the lib) is the one with commas, the no commas is the non-standard version! I'll make some edits to this in https://github.com/catppuccin/userstyles/pull/1846 but just wanted to note it ahah.

PR #1817, opened 2025-08-03
fix(desmos): style bottom of share snapshot dialog

gradylink:
## 🔧 What does this fix? 🔧

- Themes the bottom of the share dialog for snapshots
- Fixes the theming for checkboxes (there appear to have been some changes since they were originally themed)

## 🗒 Checklist 🗒

### DIFF
--- a/styles/desmos/catppuccin.user.less
+++ b/styles/desmos/catppuccin.user.less
@@ -488,7 +488,10 @@
       border-color: @base !important;
     }
 
-    .dcg-gray-link, .dcg-centered-top-link, .dcg-cancel-while-loading {
+    .dcg-gray-link,
+    .dcg-centered-top-link,
+    .dcg-cancel-while-loading,
+    .dcg-checkbox.dcg-hovered .dcg-checkbox__box.dcg-icon-check::before {
       color: @accent !important;
     }
 
@@ -551,7 +554,7 @@
     .dcg-row:nth-of-type(2) .dcg-cell,
     table th,
     table td,
-    .footer,
+    dcg-checkbox .footer,
     .dcg-add-inference-wizard--create-button-wrapper,
     .dcg-small-screen-menu__logout-container,
     .dcg-product-filter__clear-filter-container {
@@ -717,7 +720,8 @@
       border-bottom-color: @surface2 !important;
     }
 
-    .dcg-checkbox__box.dcg-icon-check.dcg-checked::before {
+    .dcg-checkbox__box.dcg-icon-check.dcg-checked::before,
+    .dcg-checkbox.dcg-hovered .dcg-checkbox__box.dcg-icon-check::before {
       text-shadow: none;
       paint-order: stroke fill;
       -webkit-text-stroke: 4px @text;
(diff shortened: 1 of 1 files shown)


uncenter:
If it works for you, it works for you - but I just want to say I think adding many unrelated classes to one block to theme them the same way gets messy quickly. I prefer to have separate sections for each component and widget and page, and if I have multiple declarations for `border-color: text` so be it.

Also, I'll leave this open for a little bit to accumulate any more fixes.

gradylink:
> If it works for you, it works for you - but I just want to say I think adding many unrelated classes to one block to theme them the same way gets messy quickly. I prefer to have separate sections for each component and widget and page, and if I have multiple declarations for border-color: text so be it.

I prefer to have a single block because there are a lot of similar components (despite having wildly different names) with desmos so separating it by page is enough for me.

> Also, I'll leave this open for a little bit to accumulate any more fixes.

Good idea

styles/desmos/catppuccin.user.less
```diff
@@ -714,6 +722,27 @@
...
+      i::before {
+        color: @base !important;
+      }
+    }
+
+    @layer {
+      .dcg-notice * {
+        color: @base !important;
+      }
+    }
```
uncenter, line 744: Is this layer and universal selector truly necessary?

styles/desmos/catppuccin.user.less
```diff
@@ -618,7 +624,9 @@
     .dcg-toast .dcg-undo,
     .dcg-segmented-control-container .dcg-selected i::before,
     .dcg-selected-graphs-actions__icon-btn i::before,
-    .dcg-segmented-control-btn.dcg-selected * {
+    .dcg-segmented-control-btn.dcg-selected *,
+    .dcg-top-banner,
+    .dcg-top-banner * {
```
uncenter, line 629: Should try to avoid the * selector.

styles/desmos/catppuccin.user.less
```diff
@@ -714,6 +722,27 @@
       border-bottom-color: @surface2 !important;
     }
 
+    .dcg-checkbox__box.dcg-icon-check.dcg-checked::before,
+    .dcg-checkbox.dcg-hovered .dcg-checkbox__box.dcg-icon-check::before {
+      text-shadow: none;
+      paint-order: stroke fill;
```
uncenter, line 728: Does this need to be set?

styles/desmos/catppuccin.user.less
gradylink: This code is really old, I know I wouldn't have added the layer if it wasn't necessary, Desmos uses some `!important`s internally, so this layer was likely used to override that, and I also can't really check the proper subclass names that this required because IIRC this class was only ever used in a "Checkout this new feature" kind of notice.

styles/desmos/catppuccin.user.less
gradylink: It does because I used a different method than Desmos does itself to style the checkboxes because their method causes tearing.

PR #1823, opened 2025-08-11
fix(bsky): repost colour

NekoDrone:
## 🔧 What does this fix? 🔧

The repost button on bluesky has the wrong colour because PBC changed the style and so the selector is incorrect.

Before:
Screenshot 2025-08-11 at 10 12 17 PM

After (fixed):
Screenshot 2025-08-11 at 10 11 54 PM

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [contributing guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/CONTRIBUTING.md).

PR #1825, opened 2025-08-11
fix(bsky): border-color(s) on various lines

NekoDrone:
## 🔧 What does this fix? 🔧

Various small lines around the app are unstyled.

### Discard draft modal

Before: 
Screenshot 2025-08-11 at 10 45 39 PM

After (fixed):
Screenshot 2025-08-11 at 11 07 37 PM

### On a feed:

Before:
Screenshot 2025-08-11 at 11 04 01 PM

After (fixed):
Screenshot 2025-08-11 at 11 01 59 PM

I have intentionally set the discard draft modal to be bordered as `surface2` while the rest of the app uses `surface1` as they are set as different colours by Bluesky, and so I believe we should follow suit. Happy to keep it to a single colour if that is better by convention.

## 🗒 Checklist 🗒

### DIFF
--- a/styles/bsky/catppuccin.user.less
+++ b/styles/bsky/catppuccin.user.less
@@ -383,6 +383,11 @@
       background-color: @red !important;
     }
 
+    /* Discard draft modal border */
+    [style*="border-color: rgb(54, 75, 97)"] {
+      border-color: @surface2 !important;
+    }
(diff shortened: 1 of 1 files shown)


NekoDrone:
merged with #1823

PR #1843, opened 2025-08-22
feat(lib): add lib.less standard library

uncenter:
First part of #1803.


.github/workflows/deploy.yml
uncenter, line 1: @sgoudham does this look right to you? Do you know if uploading the `lib` folder will make it available at userstyles.catppuccin.com/lib.less or at userstyles.catppuccin.com/lib/lib.less? I think we actually want the latter in this case because we will add more library modules later.

sgoudham: I can't actually remember, I had a look at python pygments but annoyingly we're using a custom gh-pages branch there.

Going by the workflow in [catppuccin/element](https://github.com/catppuccin/element/blob/70b7ee121dcef28c6c8191d60df2f88b23c89084/.github/workflows/github-pages.yml#L14-L16), I think you'll have to nest your `lib` folder if you want it to be the latter, like `lib/lib/lib.less` 

It'd be pretty ugly in source so I'd probably create the nested directory in CI itself.

uncenter:
Removed the deployment part of this. We will deal with that in the Starlight pull request.

PR #1845, opened 2025-08-23
feat(facebook): init

kuchingneko28:
## 🎉 Theme for Facebook🎉

Facebook is an social media and social networking service owned by Meta
## 🗒 Checklist 🗒

- [ ] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [ ] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [ ] I have ensured that the new directory is in **lower-kebab-case**.
  - [ ] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [ ] I have made sure to update the [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml) file with information about the new userstyle.
- [ ] I have included the following files:
  - [ ] `catppuccin.user.less` - all the CSS for the userstyle, based on the template.

### DIFF
--- a/styles/facebook/catppuccin.user.less
+++ b/styles/facebook/catppuccin.user.less
@@ -199,10 +199,10 @@
     --primary-button-background: @accent;
     --primary-button-text: @base;
     --primary-button-icon: @base;
-    --primary-button-background-on-color: @base;
-    --primary-button-background-on-media: @base;
-    --primary-button-text-on-media: @crust;
-    --primary-button-icon-on-color: @crust;
+    --primary-button-background-on-color: var(--always-white);
+    --primary-button-background-on-media: var(--always-white);
+    --primary-button-text-on-media: var(--always-black);
+    --primary-button-icon-on-color: var(--always-black);
     --primary-button-pressed: lighten(@accent, 22%);
     --primary-deemphasized-button-pressed: fade(@accent, 20%);
     --primary-deemphasized-button-pressed-overlay: fade(@accent, 15%);
@@ -259,15 +259,16 @@
     --story-seen: @overlay1;
 
     /* neutral */
-    --media-inner-border: fade(@overlay0, 20%);
-    --media-outer-border: @base;
+    --media-inner-border: color-mix(in srgb, var(--always-white) 5%, transparent);
+    --media-outer-border: color-mix(in srgb, var(--surface-background) 94%, var(--always-white) 6%);
     --overlay-on-media: fade(@crust, 80%);
-    --primary-text-on-media: @base;
-    --secondary-text-on-media: fade(@base, 90%);
-    --placeholder-text-on-media: fade(@base, 80%);
-    --link-on-media: @base;
+    --primary-text-on-media: var(--always-white);
+    --secondary-text-on-media: color-mix(in srgb, var(--always-white) 90%, transparent);
+    --placeholder-text-on-media: color-mix(in srgb, var(--always-white) 80%, transparent);
+    --link-on-media: var(--always-white);
     --primary-icon-on-media: var(--primary-text-on-media);
-    --secondary-icon-on-media: fade(@base, 80%);
+    --secondary-icon-on-media: color-mix(in srgb, var(--always-white) 80%, transparent);
+
 
     /* Scrollbar */
     --scroll-thumb: @overlay1 !important;
@@ -312,8 +313,9 @@
     /* Progress (neutral & on-media) */
     --progress-ring-neutral-background: fade(@text, 20%);
     --progress-ring-neutral-foreground: @text;
-    --progress-ring-on-media-background: fade(@base, 20%);
-    --progress-ring-on-media-foreground: @base;
+    --progress-ring-on-media-background: color-mix(in srgb, var(--always-white) 20%, transparent);
+
+    --progress-ring-on-media-foreground: var(--always-white);
 
     /* Messenger extras */
     --mwp-header-button-color: @accent;
@@ -387,7 +389,7 @@
     --filter-disabled-icon: @overlay2-filter;
     --filter-secondary-icon: @overlay1-filter;
     --filter-placeholder-icon: @overlay1-filter;
-    --filter-secondary-button-icon-on-media: @base-filter;
+    --filter-secondary-button-icon-on-media: var(--filter-always-white);
 
     /* Optional */
     --link-color-visited: @mauve;


styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    input,
+    textarea {
+      &::placeholder {
+        color: @subtext0 !important;
+      }
+    }
+
+    /* Start styling your website here: */
```
uncenter:
```suggestion
```

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,418 @@
...
+      --reaction-support: @yellow !important;
+
+      /* Stories / on-media text */
+      --story-unseen: @accent !important;
+      --story-seen: @overlay1 !important;
+      --primary-text-on-media: #fff !important;
+      --secondary-text-on-media: fade(#fff, 88%) !important;
+      --placeholder-text-on-media: fade(#fff, 60%) !important;
+      --primary-icon-on-media: #fff !important;
+      --link-on-media: #fff !important;
```
uncenter: Need to be themed.

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,418 @@
...
+      --filter-blue-link-icon: invert(100%) !important;
```
uncenter: re: `invert(100%)` and filters, see https://github.com/catppuccin/userstyles/blob/main/docs/guide/images-and-svgs.md#non-svg-images-or-many-img-elements-with-external-svgs

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,418 @@
...
+
+      /* Hovers / overlays */
+      --background-deemphasized: fade(@text, 10%) !important;
+      --media-hover: fade(@text, 15%) !important;
+      --media-pressed: fade(@text, 35%) !important;
+      --non-media-pressed: fade(@text, 15%) !important;
+      --media-hovered: fade(@text, 15%) !important;
+      --hover-overlay: fade(@text, 12%) !important;
+      --press-overlay: fade(@text, 20%) !important;
```
uncenter: Use the shades of the palette instead of fading the text color please.

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,418 @@
...
+    /* If you need to specify an overwrite specific to a flavor
+     *  (e.g. only when the flavor is set to latte), you can use a `when` guard clause. */
+    & when (@flavor = latte) {
+      /* ... */
+    }
+    /* You can also negate the statement above by using
+     *  a `when not` guard clause. */
+    & when not(@flavor = latte) {
+      /* ... */
+    }
```
uncenter, line 248:
```suggestion
```

uncenter (changes requested):

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,392 @@
...
+  body.__fb-dark-mode,
+  #facebook.__fb-dark-mode { #catppuccin(@darkFlavor); }
+
+  :root[data-theme="light"],
+  html[data-theme="light"],
+  body[data-theme="light"] { #catppuccin(@lightFlavor); }
+
+  :root[data-theme="dark"],
+  html[data-theme="dark"],
+  body[data-theme="dark"] { #catppuccin(@darkFlavor); }
```
uncenter: We should be able to combine the two for light mode and the two for dark mode.

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,392 @@
...
+@var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
+@var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+==/UserStyle== */
+
+@-moz-document domain("facebook.com") {
+  /* This section applies the light/dark themes automatically based on the user's system theme.
+     Use the below example CSS to start if the website you are theming doesn't have a light/dark mode toggle.
+     For an actual example of this, see https://github.com/catppuccin/userstyles/blob/39288b593278e5efa48007d5fc001292cf910672/styles/brave-search/catppuccin.user.css#L16-L29.
+  */
+
```
uncenter:
```suggestion
```

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,392 @@
...
+  /* If the website you are theming *does* have a light/dark mode toggle,
+     you can conditionally enable light/dark flavors by targeting whatever
+     attribute that website uses to indicate the theme. *The example below gives you an idea of what
+     to look for. The website you are theming may use a class-based approach (e.g. :root.theme-dark),
+     or any of the myriad of other methods like this.*
+     Use the below example CSS to start if this is the case.
+     For an actual example of this, see https://github.com/catppuccin/userstyles/blob/39288b593278e5efa48007d5fc001292cf910672/styles/youtube/catppuccin.user.css#L21-L30.
+  */
+
```
uncenter:
```suggestion
```

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    html, body, #facebook { background: var(--surface-background) !important; color: var(--primary-text) !important; }
+    .xa3vuyk { background-color:@accent !important;}  
+    
+    &,
+    :root, html, body, #facebook,
+    :root.__fb-light-mode, :root.__fb-dark-mode,
+    .__fb-light-mode, .__fb-dark-mode,
+    :root.__fb-light-mode:root, :root.__fb-dark-mode:root {
```
uncenter: Hopefully this should be redundant given the dark/light theme applicators above? They seem similar for the most part.

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,392 @@
...
+      }
+    }
+  
+
+  /* ===== Manual override ===== */
+  :root.__fb-dark-mode,  .__fb-dark-mode,  :root[data-theme="dark"] {
+    --primary-button-background: @accent !important;
+    --blue-link: @accent !important;
+    --fb-logo: @accent !important;
+    --fb-wordmark: @accent !important;
```
uncenter: Why is this necessary?

uncenter (changes requested):

uncenter:
Login screen at https://www.facebook.com/ when logged out is unthemed.

styles/facebook/catppuccin.user.less
```diff
@@ -2,7 +2,7 @@
 @name Facebook Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/facebook
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/facebook
-@version 2025.08.23
+@version 2025.09.06
```
uncenter, line 5: Don't update this. We automatically change it in CI when necessary.

styles/facebook/catppuccin.user.less
```diff
@@ -107,7 +93,8 @@
       --placeholder-text: @overlay1 !important;
       --primary-icon: @text !important;
       --secondary-icon: @overlay1 !important;
-      --disabled-icon: @overlay0 !important;
+      --disabled-icon: @overlay2 !important;
+      --filter-primary-icon: invert(40%) sepia(10%) saturate(200%) hue-rotate(180deg) brightness(95%) contrast(90%) !important;
```
uncenter: Use `@<color>-filter` to get the filter from the palette. Not sure what color this is meant to be.

styles/facebook/catppuccin.user.less
kuchingneko28: ok, sorry idk that

styles/facebook/catppuccin.user.less
```diff
@@ -68,6 +68,32 @@
...
+    }
+
+    /*  Profile  */
+
+    .__fb-light-mode.x1r1pt67 {
+      background: @mantle !important;
+    }
+    .__fb-light-mode.x1r1pt67 span {
+      color: @text !important;
+    }
```
uncenter, line 134: Can we avoid using these autogenerated classnames at all? Aria roles or ids or test attributes are all more resistant to changes than these.

kuchingneko28: ok, i try find better options

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,339 @@
...
+    ._4jy2 {
+      background-color: @surface2 !important;
+    }
+    ._6ltj a {
+      color: @text !important;
+    }
+
+
+     &,
+    :root {
```
uncenter:
```suggestion
```

I don't think we need this wrapper? `:root` is either in the applicator selector (`:root[data-theme="light"]` or `.__fb-light-mode:root`).

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,339 @@
...
+    }
+    ._8icy ._6lth {
+      background-color: @accent !important;
+    }
+    ._4jy2 {
+      background-color: @surface2 !important;
+    }
+    ._6ltj a {
+      color: @text !important;
+    }
```
uncenter: Do we need to use these autogenerated classnames?

kuchingneko28: Yes, login page doesn't use variables they using that class name

styles/facebook/catppuccin.user.less
kuchingneko28: can i change like this ? 
```less
@media (prefers-color-scheme: light) {
  :root {
    #catppuccin(@lightFlavor);
  }
}

  :root {
    #catppuccin(@darkFlavor);
  }
}
```

so i dont use this anymore
```less
:root[data-theme="light"],
.__fb-light-mode:root,
.__fb-light-mode {
      #catppuccin(@lightFlavor);
}
    
:root[data-theme="dark"],
.__fb-dark-mode:root,
.__fb-dark-mode {
      #catppuccin(@darkFlavor);
}
```

styles/facebook/catppuccin.user.less
uncenter: Can we use other selectors such as attribute selectors to avoid this?

kuchingneko28: i can't find other selectors, the div only have class and has no attribute

WalkQuackBack:
Just been looking through testing using this, a few things I found that were unthemed:
image
Tooltip on Post Audience

image
Button press highlight

image
Add to your post Icons

image

Skeleton loaders on the emoji picker

image
Main feed page icons

Messenger messages

kuchingneko28:
Thanks for testing it, I'll fix that soon

teohhanhui:
Any plans to also do messenger.com?

kuchingneko28:
Ahh i didnt know that, i try remove that later> Why is `!important` necessary everywhere? We can usually avoid it by improving the selectors that the variables are applied to. E.g. if the variables are applied by Facebook to the body but we apply them to the root than that would cause specificity issues that while !important does solve, could be better fixed by just targetting the variable declaration selector better.

kuchingneko28:
> Why is `!important` necessary everywhere? We can usually avoid it by improving the selectors that the variables are applied to. E.g. if the variables are applied by Facebook to the body but we apply them to the root than that would cause specificity issues that while !important does solve, could be better fixed by just targetting the variable declaration selector better.

Ahh i didnt know that, i try remove that later

kuchingneko28:
> Any plans to also do messenger.com?

I don't know, but maybe i'll try

WalkQuackBack:
> Why is `!important` necessary everywhere? We can usually avoid it by improving the selectors that the variables are applied to. E.g. if the variables are applied by Facebook to the body but we apply them to the root than that would cause specificity issues that while !important does solve, could be better fixed by just targetting the variable declaration selector better.

Reviewing the usage, in many places even though they are applied to __fb_light_mode, they are in an inline `style` attribute which needs !important to override.

kuchingneko28:
> Why is `!important` necessary everywhere? We can usually avoid it by improving the selectors that the variables are applied to. E.g. if the variables are applied by Facebook to the body but we apply them to the root than that would cause specificity issues that while !important does solve, could be better fixed by just targetting the variable declaration selector better.

I just updated the selector, and it works without using `!important` now. 😺

kmk142789 (approved):

styles/facebook/catppuccin.user.less
```diff
@@ -0,0 +1,475 @@
...
+    --messenger-card-background: @mantle;
+    --messenger-reply-background: @surface0;
+
+    /* Text & Icons */
+    --primary-text: @text;
+    --secondary-text: @subtext0;
+    --section-header-text: @subtext1;
+    --placeholder-text: @overlay1;
+    --primary-icon: @text !important;
+    --secondary-icon: @overlay1;
```
uncenter, line 161: These are the icons used at the tabs at the top of the page when signed in, they seem too dim to me.
```suggestion
    --secondary-icon: @overlay2;
```

uncenter (changes requested):
On the login/sign up page:

CleanShot 2025-10-06 at 17 03 28

1. There is this subtitle text that is invisible unless selected underneath the brand mark.
2. This button doesn't have enough contrast. Make the text color `base` when it is on top of blue/accent. Additionally, the background color on :hover is not set.
3. This button also doesn't have enough contrast for the text, should use `base` here as well. And the :hover background color is set, but it appears to be the same as the normal background so you can't tell if you are hovering over it or not. We should slightly modify darken/lighten the green background when hovered.

This eye/password reveal icon in the login input field is not themed, you can use a text filter (https://userstyles.catppuccin.com/contributing/guides/images-and-svgs/#monochrome-img-elements):

CleanShot 2025-10-06 at 17 07 03

If you click forgot password, the resulting page is mostly unthemed:

CleanShot 2025-10-06 at 17 08 06

Creating a new account flow is unthemed:

CleanShot 2025-10-06 at 17 08 30

And so is the page at https://www.facebook.com/login/ (which is a login screen seemingly identical to https://www.facebook.com when logged out, but unthemed nevertheless).

These media + other icons in the post composer and at the top of the home feed could be themed with filters as well, since they are monochrome:

CleanShot 2025-10-06 at 17 12 23

CleanShot 2025-10-06 at 17 12 30

kuchingneko28:
Hi sorry for the late reply. I’m pretty busy right now, but I’ll get the updates done as soon as I can. Thanks

WalkQuackBack:
It has been several months with no activity, so I'll close the pull request for now and you are always free to reland this!

Issue #1838, opened 2025-08-14
accent-colored boxes around custom emoji


oliviafloof:
### Is there an existing issue outlining your problem?

- [x] I am using the latest version of the provided userstyle.
- [x] I have searched the existing issues and they do not solve my problem.

### What userstyles are you seeing the problem on?

lbl:cinny

### Describe your problem.

in cinny 4.9.0 there are accent-colored boxes around custom emoji
they should not be there

### Attach screenshots.

<img width="260" height="81" alt="Image" src="https://github.com/user-attachments/assets/195dcc29-498a-4ed6-95be-46601d5cb141" />

### What browser(s) are you seeing the problem on?

firefox 141.0.3

PR #1861, opened 2025-08-30
fix(cinny): color emojis and search results properly

jn-sena:
## 🔧 What does this fix? 🔧

- removed the background color in emojis
- made the background colors of message search results work

| old | now |
| --- | --- |
| image | image |
| image | image |

## 🗒 Checklist 🗒

### DIFF
--- a/styles/cinny/catppuccin.user.less
+++ b/styles/cinny/catppuccin.user.less
@@ -169,7 +169,7 @@
     --ic-danger-high: @maroon;
     --ic-danger-normal: @maroon;
 
-    /* Users */
+    /* Legacy user colors */
     --mx-uc-1: @blue;
     --mx-uc-2: @pink;
     --mx-uc-3: @teal;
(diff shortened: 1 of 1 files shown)


styles/cinny/catppuccin.user.less
```diff
@@ -284,6 +215,78 @@
 
       background-color: @base;
     }
+
+    /* I don't think these actually do anything anymore, but they are
+    still defined in the app. They _might_ be removed in a future release. */
```
uncenter, line 220: If they don't do anything in the app let's remove them? It's more confusing to have them in source with no effect imo.

PRO-2684:
This should fix #1838

styles/cinny/catppuccin.user.less
jn-sena: the variables are defined and actually used in the CSS file, like for `body`'s background or `input` elements, but removing them doesn't break anything... i would like to remove them as well but am not sure. should we?

uncenter: Sure we can keep them in for now I guess.

uncenter (approved):

PR #1862, opened 2025-08-30
feat(bsky): more styles

NekoDrone:
## 🔧 What does this fix? 🔧

a ton of unstyled elements i've noticed over the last couple of weeks
- buttons on the new post modal that are disabled after you upload an image and before you hit post
- add alt text modal and other inputs elsewhere in app
- generic buttons across the app that were unstyled (like reply, repost, likes, etc.; very minute but i noticed so i made a change)
- explore tab icons (there is a stub for the 'new' trending topic badge, but as of time of writing this patch, it isn't there so I can't get the right classes)
- borders on image previews
- random ui lines (like the ones joining replies in a chain)
- icon backgrounds like those in the feeds tab
- pill buttons like those in the 'account activity notifications' modal when you click on the bell icon on an account.

## 🗒 Checklist 🗒


NekoDrone:
also @RoootTheFox if you're fine with it, i'll open another PR to add myself as maintainer for this userstyle too :D

RoootTheFox:
most of these should already be styled by existing selectors. i wonder if this is just a case of "oh the colors changed slightly" - in which case adding a bunch of new selectors wouldn't be necessary

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
       background-color: @surface0 !important;
       border-color: @overlay0 !important;
     }
+
+    /* any button that's disabled (we should start using broader overrides) */
+    button[disabled] > svg > path {
+      fill: @overlay0;
+    }
```
RoootTheFox, line 633: i like this approach

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* any default button with no styling */
+    button[role*="button"] > svg > path[fill*="hsl(211, 20%, 56%)"] {
+      fill: @overlay2;
+    }
+
+    /* exception for the like button apparently... */
+    button[role*="button"] > div > svg > path[fill*="hsl(211, 20%, 56%)"] {
+      fill: @overlay2;
+    }
```
RoootTheFox: this should already be handled by another selector (which just needs updating the hsl code)

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* sitewide text inputs */
+    /* the divs are styled separately from the input areas */
+    div[style*="border-color: rgb(46, 64, 82)"] {
+      border-color: @surface1 !important;
+    }
+
+    div[style*="border-color: rgb(32, 139, 254)"] {
+      border-color: @accent !important;
+    }
```
RoootTheFox: great :3

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* disabled button with accent color */
+    button[style*="background-color: rgb(14, 73, 137)"] {
+      background-color: fade(@accent, 30%) !important;
+    }
```
RoootTheFox: this should be part of the "some button background color (when hovering)" selector group

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* image preview borders across app */
+    [style*="border-color: rgb(74, 97, 121)"] {
+      border-color: @surface2 !important;
+    }
```
RoootTheFox, line 647: having this below the general "borders" selector would be nice (@ line ~286)

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+      div {
+        color: @crust !important;
+      }
+
+      svg > path {
+        fill: @crust;
+      }
+    }
+
+    /* TODO: new which is green but I can't get the right class cause nothing on bsky is a new trending topic atm */
```
RoootTheFox: +1

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* subtext across the site */
+    div[style*="color: rgb(147, 165, 183)"] {
+      color: @overlay2 !important;
+    }
```
RoootTheFox: "secondary text" selector @ ~101

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* random ui lines across the site */
+    div[style*="background-color: rgb(54, 75, 97)"] {
+      background-color: @surface1 !important;
+    }
```
RoootTheFox: maybe annotating this "reply borders" or something would fit more (also, i'd prefer if this was below the "border" selector)

RoootTheFox: actually, it more seems like the "thread lines" selector just lacked up-to-date colors c:

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+
+    /* round icon backgrounds (e.g. feeds) */
+    div[style*="background-color: rgb(21, 52, 85)"] {
+      background-color: @surface0 !important;
+    }
```
RoootTheFox, line 666: im *preetty* sure i had a selector for those (or another selector was doing those), but still +1

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +623,89 @@
...
+      background-color: fade(@accent, 40%) !important;
+    }
+    /* inactive, circle in pill */
+    div[style*="background-color: rgb(91, 119, 149)"]{
+      background-color: @surface2 !important;
+    }
+    /* inactive, hover */
+    div[style*="border-color: rgb(120, 142, 165)"] {
+      border-color: @overlay1 !important;
+    }
```
RoootTheFox, line 683: +1

RoootTheFox (changes requested):
looks good for the most part, but a lot of selectors can either be merged or are redundant (aka, old selectors just have to get updated with new values)

styles/bsky/catppuccin.user.less
```diff
@@ -623,6 +626,77 @@
...
+
+    /* any default button with no styling */
+    button[role*="button"] > svg > path[fill*="hsl(211, 20%, 56%)"] {
+      fill: @overlay2;
+    }
+
+    /* exception for the like button apparently... */
+    button[role*="button"] > div > svg > path[fill*="hsl(211, 20%, 56%)"] {
+      fill: @overlay2;
+    }
```
RoootTheFox: yup these are handled by another selector already, will fix

RoootTheFox (approved):

NekoDrone:
Actually, I'm looking through the userstyle and I'm wondering if we can take a more general approach to applying our overrides instead of performing these on a per-component basis.

For example, for toast notifications, (at ~371)

```diff
-    .r-17c3jg3:has(> div[dir="auto"]) {
-      background-color: @crust !important;
-
-      .r-jwli3a {
-        color: @text !important;
-      }
-      > svg > path {
-        fill: @text !important;
-      }
-    }
+    div[style*="background-color: rgb(21, 41, 63)"] {
+      background-color: fade(@accent, 30%) !important;
+    }
+    div[style*="border-color: rgb(14, 73, 137)"] {
+      border-color: @surface2 !important;
+    }
+    svg > path[fill*="hsl(211, 99%, 73.6%)"] {
+      fill: @text !important;
+    }
+    div[style*="color: rgb(121, 185, 254)"] {
+      color: @text !important;
+    }
```

Suggesting this because AFAIK on upstream, social-app is React Native and the styles simply get applied inline whenever they are defined, so we don't have to worry too much about style conflicts so long as we do specific overrides based on the colour values.

I'm noticing as well that this looks to have been the approach earlier on in the stylesheet, but after that there are a ton of component-specific overrides which I think can be improved to use this approach instead.

In fact, iirc the Less pre-processor that we use for userstyles allows us to string interpolate variables into our selectors, so the above example can be done with

```less
    @bskybg1: rgb(21, 41, 63);
    @bskyborder1: rgb(14, 73, 137);
    @bskysvgpathfill1: hsl(211, 99%, 73.6%);
    @bskytext1: rgb(121, 185, 254);

    div[style*="background-color: @{bskybg1}"] {
      background-color: fade(@accent, 30%) !important;
    }
    div[style*="border-color: @{bskyborder1}"] {
      border-color: @surface2 !important;
    }
    svg > path[fill*="@{bskysvgpathfill1}"] {
      fill: @text !important;
    }
    div[style*="color: @{bskytext1}"] {
      color: @text !important;
    }
```

NekoDrone:
also lmao looking at the toast class showed me that the toast notification border and text colour has drifted :')

RoootTheFox:
this was essentially my i initial approach, though it did look off in a lot of places. i've been thinking of writing a script of some sorts that gets the social-app color values from its build process and applies them to the userstyle. haven't yet done that though

RoootTheFox:
> also lmao looking at the toast class showed me that the toast notification border and text colour has drifted :')

yeah this is quite common. i still think that the best approach would be to ~~lobby for~~ kindly ask/pr a way to have colors be set through CSS variables instead of inline in social-app

NekoDrone:
> this was essentially my i initial approach,

would you mind if I tried the variable interpolation approach? I think I can get pretty far with it; just gotta be selective with how it overrides.

> set through CSS variables instead of inline in social-app

I don't actually know if that's possible with RN :') will need to take a look. AFAIK, RN doesn't have a way to actually do CSS variables in stylesheets because it doesn't generate actual css and instead just applies them inline at a per-component level, but I could be wrong.

PR #1866, opened 2025-08-31
feat(lib): add mediawiki lib module

uncenter:
Introduces a library module for MediaWiki sites (that are on a recent enough version of MediaWiki to have appearance/styling CSS variables). Will first be added and merged here, then userstyles for MediaWiki sites - https://github.com/catppuccin/userstyles/pull/1895, https://github.com/catppuccin/userstyles/pull/1867 - will have integration PRs merged to start using this.


ninetailedtori:
Right, having one issue with this lib rn:
When using `#mediawiki.base() !important` as might be required to apply to archwiki, given it attempts to overwrite most things locally with inlines, this requires:
```less
background-repeat: no-repeat !important;
```
otherwise certain links on some pages where it's undefined will cause the image to display per letter. Tested on archwiki, with span a.external, a.external.free, etc. However, when fixed with this, the way that the image is positioned will revert to top left as is standard behaviour for images, and when trying to reposition this, it can often cause clipping. Not too sure about a fix for the latter issue. We might need to specify a proper image placement and add padding for it, OR simply add a check that if `background-display: none` is set, we should ignore those links. Not sure if that's possible in css though o-o

ninetailedtori:
Also, we need to define an image behaviour for extiw, which hasn't been done. Current image is: url(/resources/src/mediawiki.skinning/images/external-ltr.svg?c652e) center right no-repeat

lib/mediawiki/v1.less
```diff
@@ -0,0 +1,433 @@
...
+      }
+    }
+
+    .mw-parser-output {
+      a.external {
+        @svg: escape(
+          '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" viewBox="0 0 12 12"><path fill="@{accent}" d="M6 1h5v5L8.86 3.85 4.7 8 4 7.3l4.15-4.16zM2 3h2v1H2v6h6V8h1v2a1 1 0 0 1-1 1H2a1 1 0 0 1-1-1V4a1 1 0 0 1 1-1"/></svg>'
+        );
+        background-image: url("data:image/svg+xml,@{svg}");
+      }
```
ninetailedtori, line 262:
```less
background-image: url("data:image/svg+xml,@{svg}") center right no-repeat;
```
- Otherwise, certain links on some pages where it's undefined will cause the image to display per letter when using `#mediawiki.base() !important`, OR;
- when fixed with just no-repeat, the way that the image is positioned will revert to top left as is standard behaviour for images, and when trying to reposition this, it can often cause clipping.

Not too sure about a fix for the latter issue. We should probably also manually define added padding for it, OR simply add a check that if `background-display: none` is set, we should maintain those links as `background-image: none`. Not sure if that's possible in css though? :3

ninetailedtori (changes requested):
- needs requested changes
- an added extiw image with the same center right no-repeat
- could we please add colour for .vector-pinned-container to surface0, please? This gets used quite often for stickied containers. Same with toc (table of contents) actually.

lib/mediawiki/v1.less
uncenter: Can you demonstrate that this issue with !important is present if you do the same for the Wikipedia userstyle?

ninetailedtori: Yes, I can! Lemme show:
image
image
These are with extiw, though, but using the exact same code. The issue is when a mw page has NO definition for the link icon, and unfortunately wikipedia automatically globally define their external link bg image location and padding. BUT wikipedia does NOT define extiw image characteristics, which arch DOES, but arch does NOT globally define for external links, which is why it ends up being very spotty in how it works.

uncenter: I need a reproduction URL to test that myself. I'm not seeing any difference to external links in the references section of articles when I change to `#__mediawiki.base() !important!;`. I'm guessing there is something else different about whatever page those icons are on.

uncenter: CleanShot 2026-01-03 at 10 10 47@2x

This is why the background is weird. When we add !important to the base/css variables, our external link styles override this override here. A better solution would be to split the module into `#__mediawiki.css-variables();` and other more component-organized sections so that the variables can be extra specficity but not the rest. Or we can just add the plaintext external link override to the reset.

ninetailedtori: Tested on both https://wiki.archlinux.org/title/Help:Editing and https://en.wikipedia.org/wiki/Help:Interwiki_linking btw! Hmm, specificity would be good I think!

lib/mediawiki.less
```diff
@@ -0,0 +1,433 @@
...
+      }
+    }
+
+    .mw-parser-output {
+      a.external {
+        @svg: escape(
+          '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" viewBox="0 0 12 12"><path fill="@{accent}" d="M6 1h5v5L8.86 3.85 4.7 8 4 7.3l4.15-4.16zM2 3h2v1H2v6h6V8h1v2a1 1 0 0 1-1 1H2a1 1 0 0 1-1-1V4a1 1 0 0 1 1-1"/></svg>'
+        );
+        background-image: url("data:image/svg+xml,@{svg}");
+      }
```
ninetailedtori: Do you think it's ok if we separate out the module to separate submodules? I can push a PR for that, if needs be!

WalkQuackBack (commented):
For my userstyles, I define hover and active variables at the top set to 5% and 10% and such. Should we standardize on those to make migrating over to standardized constants if we ever move over to those in the std library easier?

uncenter:
> For my userstyles, I define hover and active variables at the top set to 5% and 10% and such. Should we standardize on those to make migrating over to standardized constants if we ever move over to those in the std library easier?

I don't want to necessarily block this on that note, but I would appreciate a follow up or something if you are interested and have the bandwidth.

WalkQuackBack:
These need to be rebased to the versioned library syntax, then we may be able to merge it.

uncenter:
> These need to be rebased to the versioned library syntax, then we may be able to merge it.

Done! Ready to merge?

uncenter:
I think we can merge this as a starting point and then do some testing to make sure the dependent userstyles work correctly with the library before merging those.

WalkQuackBack (approved):

WalkQuackBack:
> I think we can merge this as a starting point and then do some testing to make sure the dependent userstyles work correctly with the library before merging those.

Sounds good.

PR #1867, opened 2025-08-31
refactor(wikipedia): use mediawiki library module

### LINKED ISSUE #1955: Wikipedia table borders are the same color as background

lbl:wikipedia

### Describe your problem.

Table background and border colors are the same which makes use of some table impossible.

### Attach screenshots.

Here is an exemple of the IPv4 header table, where border are used to define field sizes. The issue is present with all themes, will be here presented with macchiato.

With:
Image

Without:
Image

Chrome 141

### Any additional comments?

_No response_

uncenter:
Integrates the new library module for MediaWiki sites from #1866. **Must be merged after #1866 is merged into main.**


ninetailedtori:
image
missing this, oddly enough.

uncenter:
Can you please provide a reproduction URL?

ninetailedtori:
~~It seems that because the sidebar content on Help: pages are NOT defined as .infobox maybe? We can see that the content is defined as table.sidebar.sidebar-collapse.nomobile.nowraplinks.hlist.linking-and-page-manipulation with role=navigation, which is different~~
Yep, confirmed it's a missing .sidebar and .sidebar-collapse tree of formatting that needs to be done in wiki theme directly (it's outside of mw.)

styles/wikipedia/catppuccin.user.less
```diff
@@ -278,7 +117,7 @@
         background-color: @surface2;
       }
```
ninetailedtori, line 118:
```suggestion
      }

	.sidebar {
		background-color: @surface0 !important;

		.sidebar-title {
			background-color: inherit !important;
		}

		.sidebar-list-title {
			background-color: @surface1 !important;
		}
	}
```

Fixes sidebar issue, see Wikipedia contribution Help: pages, e.g. https://en.wikipedia.org/wiki/Help:Interlanguage_links

ninetailedtori (commented):
Could we add these for some of the meta-page and meta-element theming? :3

styles/wikipedia/catppuccin.user.less
```diff
@@ -54,148 +55,14 @@
...
-    background: inherit !important;
-    color: inherit !important;
-    border-color: var(--border-color-subtle) !important;
-  }
+  #__mediawiki.reset();
 
   #catppuccin(@flavor) {
     #lib.palette();
     #lib.defaults();
     #lib.css-variables();
```
uncenter, line 64: These should be targeted to the actual relevant classes, e.g. `.mw-parser-output .ombox`.

styles/wikipedia/catppuccin.user.less
uncenter: Seems like there is a specific CSS override for that exact table?

```
  html.skin-theme-clientpref-night .mw-parser-output .linking-and-page-manipulation {
    background-color:#101418;
    color:white
  }
```

ninetailedtori: Hmm, it's rather consistent for all Help pages, though? Are the others tagged the same as well?

uncenter: If you could check yourself that would be great, I'm a little spread thin at the moment.

ninetailedtori: > If you could check yourself that would be great, I'm a little spread thin at the moment.

Confirmed that we need it to be sidebar: see other pages outside of topic: https://en.wikipedia.org/wiki/Wikipedia:Project_namespace#Wikipedia_how-to_and_information_pages, https://en.wikipedia.org/wiki/Wikipedia:Policies_and_guidelines, etc.

ninetailedtori: Or, alternatively, [role="navigation"] should be more "specific" if you're worried, as well.

uncenter:
I'd appreciate it if we could avoid pasting large screenshots all over this integration thread. Let's keep questions like those to the Discord where we can have a discussion.

styles/wikipedia/catppuccin.user.less
uncenter: > Confirmed that we need it to be sidebar: see other pages outside of topic: [en.wikipedia.org/wiki/Wikipedia:Project_namespace#Wikipedia_how-to_and_information_pages](https://en.wikipedia.org/wiki/Wikipedia:Project_namespace#Wikipedia_how-to_and_information_pages), [en.wikipedia.org/wiki/Wikipedia:Policies_and_guidelines](https://en.wikipedia.org/wiki/Wikipedia:Policies_and_guidelines), etc.

Those pages use the `.help-box` class.

> Or, alternatively, [role="navigation"] should be more "specific" if you're worried, as well.

I would really like to stick to using the actual classnames on these elements if we could please. There is no need to branch out into attribute selectors here when we can look at how the original CSS does it perfectly well.

ninetailedtori:
> I'd appreciate it if we could avoid pasting large screenshots all over this integration thread. Let's keep questions like those to the Discord where we can have a discussion.

I'll move over to pingin you there then, wasn't sure which you preferred :]

styles/wikipedia/catppuccin.user.less
ninetailedtori: Changed to .help-box.

styles/wikipedia/catppuccin.user.less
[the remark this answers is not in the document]
ninetailedtori: Changed!

styles/wikipedia/catppuccin.user.less
```diff
@@ -299,7 +138,7 @@
           border-color: @surface1;
         }
       }
-      /* MediaWiki Pages */
+      /* https://www.mediawiki.org/ Pages */
```
ninetailedtori:
```suggestion
	.ombox {
		background-color: @surface0;
		border-color: @overlay0;
	}

      /* https://www.mediawiki.org/ Pages */
```

styles/wikipedia/catppuccin.user.less
ninetailedtori: NOT certain we should be specifying it with .help-box, tbh. See this: https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Linking#Interwiki_links, where "Manual of Style" sidebar is separate from linking-and-page-manipulation but still sidebar, and breaking from the same thing. Well, it's up to you.

WalkQuackBack: Unsure about this still, was an option eventually settled on?

WalkQuackBack (commented):
The icons might need another check since Codex Icons has been updated, but I don't have bandwidth to pick it up. Everything else looks good from a cursory review.

uncenter:
@ninetailedtori I'm trying to understand where we left of with your requested changes here. Could you summarize what issues, if any, are still present with this pull request's version of the Wikipedia userstyle? Are those issues fixed by #2312?

ninetailedtori:
Honestly after the whole... big merge? I gotta work through it and double check any PRs I was testing re the mediawiki lib 🥹 I'll be checking this one today to see if my comments are still relevant!

styles/wikipedia/catppuccin.user.less
ninetailedtori: I've reviewed, and there's some case like manual of style, which are only sidebar-tagged, and some that are their own, like help-box and linking-and-page-manipulation, so it's best to just do sidebar, since that includes all cases including ones that don't have their own classes! But I'm willing to accept a solution that fits better if one can be found? Just that the ones rn don't cover all the theming :3

PR #1879, opened 2025-09-02
fix(lichess): theme board, coords, and move dests

zenoix:
## Fix Unthemed Board, Coordinates, and Move Destinations

This PR creates a new stylus option to enable/disable theming of the board+coords in addition to fixing the theming of the boards and the file/rank coordinates using accents. It also changes the move destination indicators to overlay1. 

New Option:
image

Mocha Peach:
image

Latte Blue:
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/lichess/catppuccin.user.less
+++ b/styles/lichess/catppuccin.user.less
@@ -14,6 +14,7 @@
 @var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
 @var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
 @var checkbox stylePieces "Style Pieces" 1
+@var checkbox styleBoard "Style Board" 1
 ==/UserStyle== */
 
 @-moz-document domain("lichess.org") {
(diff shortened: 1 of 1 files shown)


zenoix:
The last-move styling will need to be figured out eventually. Could do what the chess.com userstyle does and make it red. But that might mean that users with red accents don't have last-move styling. We could make it so that only if the accent is red, the last-move color is something else?

uncenter:
> The last-move styling will need to be figured out eventually. Could do what the chess.com userstyle does and make it red. But that might mean that users with red accents don't have last-move styling. We could make it so that only if the accent is red, the last-move color is something else?

We can set an option for it which I think we do for Chess.com? We have four different highlight color options at least I think.

zenoix:
> We can set an option for it which I think we do for Chess.com?

Added one :)

styles/lichess/catppuccin.user.less
```diff
@@ -336,30 +361,33 @@
...
+        background: radial-gradient(
+          ellipse at center,
+          @check 0%,
+          @check 25%,
+          fade(@check, 0%) 89%,
+          fade(@check, 0%) 100%
+        );
+      }
     }
+
```
uncenter: I'm thinking we should only theme these if the rest of the board is themed. And maybe same with the analysis arrows below? I don't know who the person is who doesn't want the board themed so I can't say for sure but the contrast between Catppuccin colors overlayed on the default board isn't great.

zenoix: Fair call, I think you're right about the analysis arrows too. I'll make the changes now.

styles/lichess/catppuccin.user.less
```diff
@@ -371,16 +401,18 @@
...
+      svg.cg-shapes g {
+        &[cgHash$="green"] {
+          filter: @green-filter;
+        }
+        &[cgHash$="paleBlue"] {
+          filter: @sky-filter;
+        }
+        &[cgHash$="red"] {
+          filter: @red-filter;
+        }
```
uncenter: Is there a reason we aren't putting this up with the other styleBoard-related section?

zenoix: I thought it would've been better to place it together with the analysis related styling as the blue and red arrows only show up when doing analysis. Happy to move it if you think putting it with the other section is more appropriate

uncenter: Normally yeah but I'm not seeing any other analysis-related styling above or below here?

zenoix: You're right lol. I swear I had the eval bar theming right below it but I guess not. I'll move the analysis arrow stuff to be with all the other styleBoard stuff.

zenoix: I've moved that section now

PR #1887, opened 2025-09-06
refactor(google-drive): integrate m3 lib

WalkQuackBack:
Integrate M3 library module into Google Drive.


styles/google-drive/catppuccin.user.less
```diff
@@ -45,241 +56,132 @@
...
-
-    & when not(@flavor = latte) {
-      --dt-scrim: fade(@base, 50%);
-      --dt-scrim-2x: fade(@base, 50%);
-      --gm3-sys-color-scrim: @base;
-      --gm3-sys-color-scrim-rgb: #lib.rgbify(@base)[];
-      --gm3-sys-color-shadow: @base;
-      --gm3-sys-color-shadow-rgb: #lib.rgbify(@base)[];
-    }
+    --dt-blue-fill: #a1c9ff;
```
WalkQuackBack: Not sure where these are used, but not sure what the best course of action to map these would be.

styles/google-drive/catppuccin.user.less
```diff
@@ -606,21 +519,6 @@
     .XKSfm-Sx9Kwc-bN97Pc {
       color-scheme: auto; /* unset doesn't work? */
     }
-    [role="tooltip"],
```
WalkQuackBack, line 609: This is to conform with Google M3 style guide, which states inverse surface should be used for tooltips.

styles/google-drive/catppuccin.user.less
uncenter: If we don't know where they are used we can remove them. Or if you want to test with them we could set them to `red` (NOT `@red`) so that they stand out boldly when you come across them.

WalkQuackBack: Found out where they're used: will theme them in next pass

r58Playz (approved):
Looks good, though only the video player's loading progress is themed for me. The other video player elements look like they're hardcoded

WalkQuackBack:
> Looks good, though only the video player's loading progress is themed for me. The other video player elements look like they're hardcoded

Ah, missed that. I'll see if I can theme them without using too many generated classes.

WalkQuackBack:
Themed video player. Accidentally hit enter too early on the keyboard, whoops.

styles/google-drive/catppuccin.user.less
```diff
@@ -493,6 +493,61 @@
...
+      .Z3tdfe::before {
+        background-color: fade(@ae-media-theme[@base], 30%);
+      }
+      /* Slider value indicator */
+      .E6snnb, .dJ5x8c {
+        --gm3-slider-value-indicator-label-label-text-color: @text;
+        --gm3-slider-value-indicator-container-color: fade(@ae-media-theme[@surface0], 60%);
+      }
+    }
+
```
uncenter: Must we use autogenerated classnames...

WalkQuackBack: I suppose there's some data- attributes that could be used to select the elements

styles/google-drive/catppuccin.user.less
```diff
@@ -276,33 +282,18 @@
...
-        }
-        & when (@flavor = macchiato) {
-          filter: brightness(0) invert(13%) sepia(26%) saturate(823%)
-            hue-rotate(195deg) brightness(90%) contrast(92%);
-        }
-        & when (@flavor = mocha) {
-          filter: brightness(0) invert(10%) sepia(36%) saturate(650%)
-            hue-rotate(201deg) brightness(90%) contrast(93%);
-        }
+        filter: @catppuccin-filters[@@flavor][@@accentColor];
```
uncenter:
```suggestion
        filter: @accent-filter;
```

Same with the rest. You aren't meant to be using `@catppuccin-filters`.

styles/google-drive/catppuccin.user.less
WalkQuackBack: Doesn't seem like there's a better way to select the elements that doesn't overscope and select other things unintentionally...

WalkQuackBack:
Anything that still needs changing?

uncenter:
> Anything that still needs changing?

Lint failure is all. Then this is good to go for after M3 lib module is merged.

uncenter:
Still need to fix that (autofixable) lint error :)

PR #1888, opened 2025-09-06
feat(tldraw): rewrite

NekoDrone:
## 🔧 What does this fix? 🔧

some time within the last couple of weeks, tldraw had a full rewrite of their frontend style solution. this has caused the entire tldraw userstyle to be fully broken, requiring a rewrite.

the new version is thankfully much easier to style with variables and requires less inline overriding (there is still some, unfortunately).

the same rule applies, where exported images will remain unstyled.

for the light/dark variants of the same colour (red, green, violet), i have opted to darken the colours instead of lightening as it better conforms with the exported non-styled images and provides better visual hierarchy.

## 🗒 Checklist 🗒


styles/tldraw/catppuccin.user.less
```diff
@@ -21,368 +21,326 @@
...
-      color: var(--color-light-blue) !important;
-    }
-    [title="Color — Violet"] {
-      color: var(--color-violet) !important;
-    }
-    [title="Color — Light violet"] {
-      color: var(--color-light-violet) !important;
+    @violet: darken(@mauve, 20%);
+    @dark-green: darken(@green, 15%);
+    @dark-red: darken(@red, 10%);
```
uncenter, line 74: Can't we just use `maroon` as normal red and `red` as dark red.

NekoDrone: not enough visual contrast i think. previous userstyle used `@red` as the darker red then `lighten(@red, 5%)` as the lighter ver but i think this version is clearer on the separation.

uncenter (changes requested):
We should set `--tl-color-grid` for when Preferences > Show grid is enabled.

uncenter (changes requested):
Looks like solid filled shapes aren't themed:

CleanShot 2025-09-06 at 11 26 04

CleanShot 2025-09-06 at 11 26 01

styles/tldraw/catppuccin.user.less
```diff
@@ -21,368 +21,327 @@
...
-      background: linear-gradient(
-        180deg,
-        @surface0 0%,
-        var(--color-muted-2) 100%
-      );
-    }
-    .tlui-people-menu__avatar {
-      background-color: @green !important;
-    }
+    --tl-color-background: @crust;
```
uncenter: I think we should set the primary drawing background to base, and then the UI overlays and sidebar to like mantle and/or surface?

styles/tldraw/catppuccin.user.less
```diff
@@ -173,6 +173,54 @@
...
+          fill: fade(@green, 20%);
+        }
+
+        path[fill="#3c2b2b"] {
+          fill: fade(@red, 20%);
+        }
+
+        path[fill="#382726"] {
+          fill: fade(@dark-red, 20%);
+        }
```
uncenter: Would be helpful to add a comment here detailing what this addition themes since it is unclear from the code itself.

NekoDrone: yes, also realised that each of these need a light mode variant too, including the ones elsewhere because we're discriminating based on specific colours :')

styles/tldraw/catppuccin.user.less
NekoDrone: hmm, I'm not sure which is better because of this section of the UI.

Screenshot 2025-09-07 at 12 57 02 AM

if i lighten the canvas to `base`, then by hierarchy it would need to look like either of the two below. i'm not sure which one is better overall.

Screenshot 2025-09-07 at 12 59 32 AM
Screenshot 2025-09-07 at 12 59 52 AM

for both cases, other UI elements look like this

Screenshot 2025-09-07 at 1 03 46 AM

Screenshot 2025-09-07 at 1 04 02 AM

Screenshot 2025-09-07 at 1 03 48 AM

Screenshot 2025-09-07 at 1 04 05 AM

---

my other suggestion is to set the canvas to `mantle`, which would result in a consistent hierarchy

Screenshot 2025-09-07 at 1 07 10 AM
Screenshot 2025-09-07 at 1 07 12 AM
Screenshot 2025-09-07 at 1 07 15 AM

uncenter: Asked Hammy in Discord (#userstyles) and he liked your mantle approach. Let's do that one!

styles/tldraw/catppuccin.user.less
```diff
@@ -21,368 +21,452 @@
...
+
   .tl-theme__light {
     #catppuccin(@lightFlavor);
   }
 
+  /* for clerk components in modal */
+  /* force latte because clerk on tldraw doesn't have dark mode set */
+  .cl-modalContent {
+    #catppuccin(@lightFlavor);
+  }
```
uncenter: We could just use media queries here actually.

styles/tldraw/catppuccin.user.less
```diff
@@ -21,368 +21,452 @@
...
+    --tl-color-text-2: @subtext1;
+    --tl-color-text-3: @subtext0;
+    --tl-color-selected-contrast: @crust;
+    --tl-color-snap: @red;
+    --tl-color-selection-stroke: @accent;
+    --tl-color-divider: @surface1;
+    --tl-color-grid: @overlay1;
+    --tla-color-primary: @accent;
+    --tla-color-primary-hover: fade(@accent, 80%);
+    --tla-color-sidebar: @surface0;
```
uncenter, line 64: I'd argue we should make this `base`. Looks too bright for a sidebar.
```suggestion
    --tla-color-sidebar: @base;
```

NekoDrone: yeah this was one of my concerns and is why i put `crust` as the main canvas background before.

if the sidebar is set to base, you get this:

sidebar with `base`:
Screenshot 2025-09-07 at 3 25 29 PM

whereas the current style is this:
Screenshot 2025-09-07 at 3 26 45 PM

and the original unstyled site looks like this:
Screenshot 2025-09-07 at 3 28 16 PM

styles/tldraw/catppuccin.user.less
NekoDrone: wdym? like this?

```less
  .cl-modalContent {
    @media (prefers-color-scheme: light) {
      #catppuccin(@lightFlavor);
    }

      #catppuccin(@lightFlavor);
    }
  }
```

for the clerk components?

styles/tldraw/catppuccin.user.less
uncenter: Why can't we make the percent thing a different color as well, make that surface0?

NekoDrone: not at my pc atm but iirc it's linked to `--tl-color-low`, and so if we set that to `surface0`, it clashes elsewhere.

styles/tldraw/catppuccin.user.less
uncenter: Mhm that's what I was thinking! Do you think that makes sense or?

styles/tldraw/catppuccin.user.less
uncenter: Alright we can leave it as is.

uncenter:
Does this fix everything in https://github.com/catppuccin/userstyles/issues?q=sort%3Aupdated-desc%20state%3Aopen%20label%3A%22tldraw%22%20is%3Aissue?

NekoDrone:
> Does this fix everything in https://github.com/catppuccin/userstyles/issues?q=sort%3Aupdated-desc%20state%3Aopen%20label%3A%22tldraw%22%20is%3Aissue?

yes. all the previously reported issues were there because of frontend changes on tldraw's end. this PR should close all prior tldraw issues.

styles/tldraw/catppuccin.user.less
NekoDrone: yeah now that I think about it, it'll be better to be explicit for future maintenance, will make the change.

uncenter: Actually, since when you are signed out tldraw forces light mode and doesn't follow system theme, we can just keep forcing light mode. Might look weird to have mocha/dark login modal over latte/light tldraw?

NekoDrone: when signed out it should still let you set the theme through preferences > theme tho. if it's via incognito/private, tldraw will use light. either way, for the clerk sign in modal, since it's light mode we should force latte anyway (or whichever the user sets as the light flavour i suppose) and either approach works.

i would do the media query doubling just to be explicit for the future in that case.

uncenter: Ah your right, preferences do apply but it defaults to light I think. You could do `:root:has(.tl-container.tl-theme__dark) .cl-modalContent` and `:root:has(.tl-container.tl-theme__light) .cl-modalContent` instead of a media query then.

styles/tldraw/catppuccin.user.less
```diff
@@ -21,368 +21,456 @@
...
+
   .tl-theme__light {
     #catppuccin(@lightFlavor);
   }
 
+  /* for clerk components in modal */
+  /* force latte because clerk on tldraw doesn't have dark mode set */
+  :root:has(.tl-container.tl-theme__dark) .cl-modalContent {
+    #catppuccin(@lightFlavor);
+  }
```
uncenter:
```suggestion
  :root:has(.tl-container.tl-theme__dark) .cl-modalContent {
    #catppuccin(@darkFlavor);
  }
```

We don't need to force light here though, we can respect the preference and make this dark.

NekoDrone: The modal has a png image that's black and can't be re-coloured. Also the Clerk components themselves are only light mode. No dark mode variant exists.

uncenter: I can get it looking pretty good...

CleanShot 2025-09-10 at 09 30 14

> No dark mode variant exists.

If we theme them properly they will look correct in any flavor, which means whether the original has a dark mode variant or not to begin with is irrelevant.

NekoDrone: Huh. This must have been a recent change because the tldraw logo in the middle was a png before :o

uncenter: I'll push my changes :)

uncenter: The trick is the source still has a PNG but we can overlay it with an SVG from another source.

PR #1896, opened 2025-09-08
docs(guides/images-and-svgs): monochrome images, inlined svg data urls

uncenter:
Moves the formerly titled "Non-SVG images or many `` elements with external SVGs" section to the top, now titled "Monochrome `` elements". Since the filter palette and variables are included by default with the standard library, this is now by far the easiest and simplest way for many usecases.

In the "SVGs as `background-image`s" section, I've split the obtaining of the SVG contents into two subsections depending on the original format. This includes a new subsection on "Inlined SVG data URL"s.

An additional note about single quotes when pasting SVG contents, some wording changes.

### DIFF
--- a/docs/src/content/docs/contributing/guides/images-and-svgs.md
+++ b/docs/src/content/docs/contributing/guides/images-and-svgs.md
@@ -58,7 +58,7 @@ Once you have obtained the SVG contents, paste them in between the single quotes
 ```
 
 > [!NOTE]
-> Make sure the quotes in `escape('')` are single quotes, not double quotes (`escape("")`), before pasting the SVG contents. Using double quotes will not work as double quotes are also used in the contents of SVGs.
+> Ensure `escape()` uses single quotes, not double quotes, before pasting the SVG contents — `escape('')`, not `escape("")`.
 
 > [!WARNING]
 > The error `Invalid % without number` appears until an interpolated color is added in the SVG contents. This is covered in the following section.
--- a/docs/src/content/docs/contributing/guides/images-and-svgs.md
+++ b/docs/src/content/docs/contributing/guides/images-and-svgs.md
@@ -43,7 +43,11 @@ Below is an example of what a rule for an inlined SVG data URL could look like.
 }
 ```
 
-The contents past `data:image/svg+xml,` are almost gibberish. This is because they are URL encoded. To get the actual SVG contents, you'll need to URL decode the string. This can be done with the JavaScript builtin function `decodeURIComponent`; run `node -e 'console.log(decodeURIComponent("..."))'`, inserting your URL encoded SVG contents in between the double quotes in place of the `...`. Alternatively, use an online tool such as https://www.urldecoder.io/.
+The text after `data:image/svg+xml,` is URL encoded. To get the contents of the SVG, the string needs to be URL decoded.
+
+There are two ways to do this:
+- Using the JavaScript built-in function `decodeURIComponent`. You can run `node -e 'console.log(decodeURIComponent("..."))'`, with the URL encoded SVG contents placed between the double quotes in place of `...`.
+- Use an online tool such as [https://www.urldecoder.io/](URLDecoder).
 
 ---
 
--- a/docs/src/content/docs/contributing/guides/images-and-svgs.md
+++ b/docs/src/content/docs/contributing/guides/images-and-svgs.md
@@ -7,7 +7,7 @@ sidebar:
 
 ## Monochrome `<img>` elements
 
-If you encounter an `<img>` element with bitonal contents, the easiest way to theme it is to apply a CSS color filter. The `@<color>-filter` filters can be used to theme Catppuccin colours.
+If you encounter an `<img>` element with single colored contents, the easiest way to theme it is to apply a CSS color filter. The `@<color>-filter` filters can be used to theme Catppuccin colours.
 
 For example:
 
--- a/docs/src/content/docs/contributing/guides/images-and-svgs.md
+++ b/docs/src/content/docs/contributing/guides/images-and-svgs.md
@@ -61,7 +61,7 @@ Once you have obtained the SVG contents, paste them in between the single quotes
 > Make sure the quotes in `escape('')` are single quotes, not double quotes (`escape("")`), before pasting the SVG contents. Using double quotes will not work as double quotes are also used in the contents of SVGs.
 
 > [!WARNING]
-> The `Invalid % without number` error may appear until an interpolated color is added in the SVG contents (as is detailed below).
+> The error `Invalid % without number` appears until an interpolated color is added in the SVG contents. This is covered in the following section.
 
 Now, replace any colors in the SVG with their respective Catppuccin variants. For example, take the following SVG icon for Twitter:
 
--- a/docs/src/content/docs/contributing/guides/images-and-svgs.md
+++ b/docs/src/content/docs/contributing/guides/images-and-svgs.md
@@ -31,7 +31,7 @@ Below is an example of what a rule for an external SVG URL could look like.
 }
 ```
 
-The easiest way to theme external SVGs is to visit the URL of the SVG and paste its contents.
+To obtain the contents of this remote SVG resource, you can either: visit the URL of the SVG, right click and select **View Page Source**, and copy its contents from the resulting page; or, run `curl "https://example.org/assets/xyz.svg"` and copy the contents from the terminal output (copy the contents automatically with `curl "https://example.org/assets/xyz.svg" | pbcopy` on macOS).
 
 ### Inlined SVG data URL
 
--- a/docs/src/content/docs/contributing/guides/images-and-svgs.md
+++ b/docs/src/content/docs/contributing/guides/images-and-svgs.md
@@ -31,7 +31,12 @@ Below is an example of what a rule for an external SVG URL could look like.
 }
 ```
 
-To obtain the contents of this remote SVG resource, you can either: visit the URL of the SVG, right click and select **View Page Source**, and copy its contents from the resulting page; or, run `curl "https://example.org/assets/xyz.svg"` and copy the contents from the terminal output (copy the contents automatically with `curl "https://example.org/assets/xyz.svg" | pbcopy` on macOS).
+To obtain the contents of this remote SVG resource, you can either:
+- Visit the URL of the SVG, right click and select **View Page Source**, and copy its contents from the resulting page.
+- Run `curl "https://example.org/assets/xyz.svg"` and copy the contents from the terminal output.
+
+> [!TIP]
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,63 @@ sidebar:
...
 ---
 
+## Monochrome `<img>` elements
+
+If you encounter an `<img>` element with monochrome contents, the easiest approach is to apply a CSS color filter. You can use the `@<color>-filter` variables provided out of the box.
+
+For example:
+
+```css
+img.twitter-icon {
```
WalkQuackBack, line 15: Twitter seems like a bit of a bad example considering it's no longer named that— maybe this could be something else like error-icon?

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,63 @@ sidebar:
...
+Below is an example of what a rule for an external SVG URL could look like.
 
 ```css
 .xyz {
   background-image: url("https://example.org/assets/xyz.svg");
 }
 ```
 
-The easiest way to theme external SVGs is to visit the URL of the SVG and paste its contents in between the single quotes of the `escape('')` section of the following template:
+To obtain the contents of this remote SVG resource, you can either: visit the URL of the SVG, right click and select **View Page Source**, and copy its contents from the resulting page; or, run `curl "https://example.org/assets/xyz.svg"` and copy the contents from the terminal output (copy the contents automatically with `curl "https://example.org/assets/xyz.svg" | pbcopy` on macOS).
```
WalkQuackBack: This feels like it would be better split up into bullet points, or two sentences, and the part in brackets could also be split into another sentence for better scanability. Not sure how to suggest changes, but I can suggest some rewording if you don't mind.

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,63 @@ sidebar:
...
+
+Below is an example of what a rule for an inlined SVG data URL could look like.
 
+```css
+.xyz {
+  background-image: url("data:image/svg+xml,%3Csvg%20viewBox%3D%220%200%2024%2024%22%20xmlns%3D%22http%3A//www.w3.org/2000/svg%22%3E%3Cpath%20fill%3D%22red%22%20d%3D%22%22/%3E%3C/svg%3E");
+}
+```
+
+The contents past `data:image/svg+xml,` are almost gibberish. This is because they are URL encoded. To get the actual SVG contents, you'll need to URL decode the string. This can be done with the JavaScript builtin function `decodeURIComponent`; run `node -e 'console.log(decodeURIComponent("..."))'`, inserting your URL encoded SVG contents in between the double quotes in place of the `...`. Alternatively, use an online tool such as https://www.urldecoder.io/.
```
WalkQuackBack: The "almost gibberish" part feels a bit wordy, and the second sentence could be reworded to not refer to you as much. Splitting this into two sentences could also help.

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,63 @@ sidebar:
...
   background-image: url("data:image/svg+xml,@{svg}");
 }
 ```
 
 > [!NOTE]
-> The `Invalid % without number` error may appear if you have not done the following step. Make sure to add/replace an interpolated color in the SVG contents (as is detailed below) for this error to go away.
+> Make sure the quotes in `escape('')` are single quotes, not double quotes (`escape("")`), before pasting the SVG contents. Using double quotes will not work as double quotes are also used in the contents of SVGs.
+
+> [!WARNING]
+> The `Invalid % without number` error may appear until an interpolated color is added in the SVG contents (as is detailed below).
```
WalkQuackBack: May should be will, this is expected to happen, so it should be written with more confidence. In addition, the brackets make it a bit hard to scan.

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,63 @@ sidebar:
...
 .xyz {
-  @svg: escape("<svg>...</svg>");
+  @svg: escape('');
   background-image: url("data:image/svg+xml,@{svg}");
 }
 ```
 
 > [!NOTE]
-> The `Invalid % without number` error may appear if you have not done the following step. Make sure to add/replace an interpolated color in the SVG contents (as is detailed below) for this error to go away.
+> Make sure the quotes in `escape('')` are single quotes, not double quotes (`escape("")`), before pasting the SVG contents. Using double quotes will not work as double quotes are also used in the contents of SVGs.
```
WalkQuackBack: The second sentence is a bit confusing to read, I think it could either be removed or worked into the first one.

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -44,9 +80,9 @@ There is only one color used, `fill="#1D9BF0"`. That hex code is a shade of blue
 }
 ```
 
-## `<img>` elements
+## External SVGs through `<img>` elements
 
-Theming an inline image is similar, but `content` is used instead of `background-image` to cover up the original image with our new one. As with the previous tip for `background-image`, you only need to update the SVG inside of the `escape('')` (see above for details).
+Theming an external SVG in an `<img>>` element is similar, but the `content` property is used instead of `background-image` to cover up the original image with our new one. As with the previous tip for `background-image`, you only need to update the SVG inside of the `escape('')` (see above for details).
```
WalkQuackBack: The > on the img is duplicated.

docs/src/content/docs/contributing/guides/images-and-svgs.md
WalkQuackBack: This should probably be addressed in another PR.

docs/src/content/docs/contributing/guides/images-and-svgs.md
WalkQuackBack:
```suggestion
The text after `data:image/svg+xml,` is URL encoded. To get the contents of the SVG, the string needs to be URL decoded.

There are two ways to do this:
- Using the JavaScript built-in function `decodeURIComponent`. You can run `node -e 'console.log(decodeURIComponent("..."))'`, with the URL encoded SVG contents placed between the double quotes in place of `...`.
- Use an online tool such as [https://www.urldecoder.io/](URLDecoder).
```

docs/src/content/docs/contributing/guides/images-and-svgs.md
WalkQuackBack:
```suggestion
> Ensure `escape()` uses single quotes, not double quotes, before pasting the SVG contents — `escape('')`, not `escape("")`.
```

docs/src/content/docs/contributing/guides/images-and-svgs.md
WalkQuackBack:
```suggestion
Theming an external SVG referenced from an `<img>` element is similar. However, the `content` property is used instead of `background-image` to replace the image.

As with the previous tip for `background-image`, only the SVG inside of the `escape('')` needs to be updated.
```

docs/src/content/docs/contributing/guides/images-and-svgs.md
WalkQuackBack:
```suggestion
> The error `Invalid % without number` appears until an interpolated color is added in the SVG contents. This is covered in the following section.
```

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,63 @@ sidebar:
   order: 2
 ---
 
+## Monochrome `<img>` elements
+
+If you encounter an `<img>` element with monochrome contents, the easiest approach is to apply a CSS color filter. You can use the `@<color>-filter` variables provided out of the box.
```
WalkQuackBack:
```suggestion
If you encounter an `<img>` element with bitonal contents, the easiest way to theme it is to apply a CSS color filter. The `@<color>-filter` filters can be used to theme Catppuccin colours.
```
Bitonal makes the most sense to me here as monochrome image tonal values will be crunched to black and white.

WalkQuackBack (changes requested):

docs/src/content/docs/contributing/guides/images-and-svgs.md
uncenter: I don't care for "X" and I like the example :)

docs/src/content/docs/contributing/guides/images-and-svgs.md
uncenter: I edited your suggestion, wdyt? Had to move a part around since it was the wrong way around.

docs/src/content/docs/contributing/guides/images-and-svgs.md
uncenter: Thanks, this is much better.

docs/src/content/docs/contributing/guides/images-and-svgs.md
uncenter: I think this section would only apply specifically to `` elements, so I wouldn't use "such as".

docs/src/content/docs/contributing/guides/images-and-svgs.md
uncenter: I've never heard the word "bitonal", maybe we just spell out "single-color image" to be unambiguous.

docs/src/content/docs/contributing/guides/images-and-svgs.md
```diff
@@ -5,27 +5,67 @@ sidebar:
   order: 2
 ---
 
+## Monochrome `<img>` elements
+
+If you encounter an `<img>` element with bitonal contents, the easiest way to theme it is to apply a CSS color filter. The `@<color>-filter` filters can be used to theme Catppuccin colours.
```
WalkQuackBack:
```suggestion
If you encounter an `<img>` element with single colored contents, the easiest way to theme it is to apply a CSS color filter. The `@<color>-filter` filters can be used to theme Catppuccin colours.
```

docs/src/content/docs/contributing/guides/images-and-svgs.md
uncenter:
```suggestion
To obtain the contents of this remote SVG resource, you can either:
- Visit the URL of the SVG, right click and select **View Page Source**, and copy its contents from the resulting page.
- Run `curl "https://example.org/assets/xyz.svg"` and copy the contents from the terminal output.

> [!TIP]
> On macOS, copy the SVG contents after `curl`-ing automatically with `| pbcopy`: `curl "https://example.org/assets/xyz.svg" | pbcopy`.
```

WalkQuackBack: Looks good to me: I'll let you commit it.

PR #1897, opened 2025-09-08
feat(lichess): unthemed UI elements

zenoix:
## 🔧 What does this fix? 🔧

This pull request aims to theme the following from #1891:

- [X] Hamburger menu icon
- [X] Hamburger menu headers/sections
- [ ] Settings menu item hovers
- [ ] Stockfish configuration buttons
- [ ] Sign in, register page (input borders, sign in button)

Screenshots:

image

## 🗒 Checklist 🗒

- [X] I have read and followed Catppuccin's [contributing guidelines](https://userstyles.catppuccin.com/contributing/).

PR #1910, opened 2025-09-12
fix(lichess): fix invisible top clock

zenoix:
## 🔧 What does this fix? 🔧

This PR fixes the invisible top clock. It does not change the color of the unthemed running clock which will be done in #1897.

Before:
Screenshot 2025-09-12 at 14 38 24

After:
Screenshot 2025-09-12 at 14 39 29

## 🗒 Checklist 🗒

### DIFF
--- a/styles/lichess/catppuccin.user.less
+++ b/styles/lichess/catppuccin.user.less
@@ -422,6 +422,9 @@
     .rclock .bar {
       background: @accent;
     }
+    .rclock-top {
+      z-index: 0;
+    }
 
     /* Game Cards */
     .game__meta,


uncenter:
When running #1897 I think my latest push might have unintentionally fixed this issue differently?

zenoix:
> When running #1897 I think my latest push might have unintentionally fixed this issue differently?

Is that so? If that's the case then this can be closed.

uncenter:
> > When running #1897 I think my latest push might have unintentionally fixed this issue differently?
> 
> Is that so? If that's the case then this can be closed.

Could you please double check that I am looking at the right thing first? 😁

zenoix:
> Could you please double check that I am looking at the right thing first? 😁

Seems like it did :))

image

PR #1927, opened 2025-09-21
fix(mdn): update variables

Koba-gh:
## 🔧 What does this fix? 🔧
I updated variables for the latest MDN.
I think there is a lot of room to improve, especially three points below.
1. I uniformly set lightness of --color-COLOR-NUM to NUM.
2. I arbitrarily set custom values for some variables and keep default for others.
3. Syntax highlighting doesn't work now.

## 🗒 Checklist 🗒

### DIFF
--- a/styles/mdn/catppuccin.user.less
+++ b/styles/mdn/catppuccin.user.less
@@ -22,141 +22,134 @@
   #lib.defaults();
   #lib.css-variables();
 
-  --text-primary: @text;
-  --text-secondary: @subtext0;
-  --text-active: @overlay1;
-  --text-inactive: fade(@subtext0, 60%);
-  --text-link: @accent;
-  --text-visited: lighten(saturate(@accent, -20%), -20%);
-  --text-invert: @base;
-  --background-primary: @base;
-  --background-secondary: @crust;
-  --background-tertiary: @mantle;
-  --background-toc-active: @crust;
-  --background-mark-yellow: fade(@yellow, 40%);
-  --background-mark-green: fade(@green, 40%);
-  --background-information: fade(@blue, 10%);
-  --background-warning: fade(@maroon, 10%);
-  --background-critical: fade(@red, 10%);
-  --background-success: fade(@green, 10%);
-  --border-primary: @surface2;
-  --border-secondary: @surface1;
-  --button-primary-default: @text;
-  --button-primary-hover: @subtext0;
-  --button-primary-active: @overlay0;
-  --button-primary-inactive: @text;
-  --button-secondary-default: @surface2;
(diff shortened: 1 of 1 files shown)


uncenter:
> Syntax highlighting doesn't work now.

Even if I paste in the contents of the Prism highlighting import directly into the userstyle I still don't see the Prism styles showing up on the `span.token`s on the pages. I have no idea why this isn't working, cc @WalkQuackBack do you have any ideas?

styles/mdn/catppuccin.user.less
```diff
@@ -22,141 +22,134 @@
...
+    --baseline-bg: hsl(hue(@blue), saturation(@blue), lightness(@surface0));
+    --baseline-engine-bg: hsl(
+      hue(@blue),
+      saturation(@blue),
+      lightness(@base)
+    );
+    --baseline-check: @blue;
+    --baseline-pill-bg: @sapphire;
+    --baseline-pill-color: @base;
+  }
```
uncenter, line 146: These don't look right as pastel palette colors on the page. Avoid changing the lightness value this much. Typically we use `lighten`/`darken` for that purpose or we can use `fade` for minor changes in many cases, e.g. `fade(@blue, 30%)`.

WalkQuackBack:
> > Syntax highlighting doesn't work now.
> 
> Even if I paste in the contents of the Prism highlighting import directly into the userstyle I still don't see the Prism styles showing up on the `span.token`s on the pages. I have no idea why this isn't working, cc @WalkQuackBack do you have any ideas?

Taking a look at this now.

WalkQuackBack:
Pretty sure this relates to the CSS being within the Web Component Shadow DOM (specifically, it uses Lit) so it's scoped from any CSS outside of it. It's within the `mdn-code-example` component. I'm not sure there's a way to theme it considering the lack of externally accessible parts or slots .

uncenter:
> Pretty sure this relates to the CSS being within the Web Component Shadow DOM (specifically, it uses Lit) so it's scoped from any CSS outside of it. It's within the `mdn-code-example` component. I'm not sure there's a way to theme it considering the lack of externally accessible parts or slots .

Thanks, I totally missed this. A shame that more things are using web components which are unstylable to us with Stylus: https://github.com/openstyles/stylus/issues/739.

styles/mdn/catppuccin.user.less
uncenter: This still needs to be changed.

uncenter:
Made the requisite changes in https://github.com/catppuccin/userstyles/pull/2157. Closing as superseded.

PR #1960, opened 2025-10-19
refactor(youtube): use single !important

WalkQuackBack:
## 🔧 What does this fix? 🔧

Use a single !important at the mixin level instead of duplicating it several times per color declaration.

## 🗒 Checklist 🗒

- [ ] I have read and followed Catppuccin's [contributing guidelines](https://userstyles.catppuccin.com/contributing/).

### DIFF
--- a/styles/youtube/catppuccin.user.less
+++ b/styles/youtube/catppuccin.user.less
@@ -1377,17 +1377,17 @@
 
 @-moz-document domain("m.youtube.com") {
   html[darker-dark-theme] {
-    #catppuccin(@darkFlavor) !important;
+    #catppuccin(@darkFlavor);
   }
   html {
-    #catppuccin(@lightFlavor) !important;
+    #catppuccin(@lightFlavor);
   }
 
   #catppuccin(@flavor) {
     #lib.palette();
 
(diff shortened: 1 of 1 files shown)


styles/youtube/catppuccin.user.less
```diff
@@ -1262,132 +1262,132 @@
...
+    #catppuccin(@lightFlavor) !important;
   }
 
   #catppuccin(@flavor) {
     #lib.palette();
 
-    background-color: @base !important;
-    color: @text !important;
+    background-color: @base;
+    color: @text;
```
uncenter: Is this really worth it?

WalkQuackBack: Maybe it impacts parse time by a bit? But I haven't profiled, so I can't verify that.

WalkQuackBack: Did some profiling, while it does increase the time spent parsing it's not by a negligible amount.

uncenter: It just adds a bunch of unnecessary !important markings when we only use it twice in the body of the mixin though.

PR #1962, opened 2025-10-24
feat(lib/mediawiki): use codex style guide

WalkQuackBack:
Use the [Wikimedia Codex color design tokens](https://doc.wikimedia.org/codex/latest/design-tokens/color.html) to base the colour bindings. 

This makes the appearance of the style more consistent with base Wikipedia.

### DIFF
--- a/lib/mediawiki.less
+++ b/lib/mediawiki.less
@@ -164,8 +164,8 @@
 
     --background-color-input-binary--checked: @accent;
 
-    --background-color-tab-list-item-framed--hover: rgba(255, 255, 255, 0.3);
-    --background-color-tab-list-item-framed--active: rgba(255, 255, 255, 0.65);
+    --background-color-tab-list-item-framed--hover: fade(@catppuccin[@latte][@base], 30%);
+    --background-color-tab-list-item-framed--active: fade(@catppuccin[@latte][@base], 65%);
 
     --border-base: 1px solid var(--border-color-base);
     --border-subtle: 1px solid var(--border-color-subtle);


WalkQuackBack:
Related to #1866

uncenter:
Can you explain more about what these changes do or fix?

WalkQuackBack:
> Can you explain more about what these changes do or fix?

Sure! Using the latest version of the library, there's cases where table backgrounds and borders use the same colour which is bad for contrast. In addition, there was a few unthemed variables that are used on other MediaWiki wikis but are part of the Codex style guide spec.

This is using [feat/mediawiki-lib-integrate](https://github.com/catppuccin/userstyles/blob/feat/mediawiki-lib-integrate/styles/wikipedia/catppuccin.user.less) modified to point to both versions of the libraries as they are not deployed yet.

First example, on [Wikipedia - Help:Sortable_tables](https://en.wikipedia.org/wiki/Help:Sortable_tables#Example)
image
_Before PR_

image
_After PR_

Second example, on [Wikipedia - Help:Introduction_to_tables_with_Wiki_Markup](https://en.wikipedia.org/w/index.php?title=Help:Introduction_to_tables_with_Wiki_Markup/All&action=edit) when signed out.
image
_Before PR_

image
_After PR_

Third example, in the footer on Wikipedia. Contrast is improved by a lot, closer to matching the style guide contrast recommendations.
image
_Before PR_

image
_After PR_

Fourth example, buttons now have a colour change when pressed.
image
_Before PR, hover_
image
_Before PR, active_

image
_After PR, hover_

image
_After PR, active_

lib/mediawiki.less
```diff
@@ -66,79 +66,150 @@
...
-    --color-progressive--focus: darken(@accent, 5%);
+    --color-progressive--hover: lighten(@accent, 5%);
+    --color-progressive--active: darken(@accent, 10%);
+    --color-progressive--focus: darken(@accent, 10%);
+    
     --color-destructive: @red;
     --color-destructive--hover: @maroon;
-    --color-destructive--active: @maroon;
+    --color-destructive--active: darken(@red, 4%);
+    --color-destructive--visited: spin(@red, 15);
```
uncenter, line 83: What does spin do?

lib/mediawiki.less
```diff
@@ -66,79 +66,150 @@
...
+
+    --background-color-content-added: mix(@blue, @base, 30%);
+    --background-color-content-removed: mix(@yellow, @base, 30%);
+
+    --background-color-transparent: transparent;
+    --background-color-backdrop-light: fade(@base, 65%);
+    --background-color-backdrop-dark: fade(@text, 65%);
+
+    --background-color-button-quiet--hover: fade(@text, 2.7%);
+    --background-color-button-quiet--active: fade(@text, 8.2%);
```
uncenter: Do we need to fade text for this?

lib/mediawiki.less
WalkQuackBack: Hue shifting. I tried using Maroon but I could hardly tell the difference between it being visited and not. I have some issues seeing colours though, so if you think that maroon contrasts well enough feel free to go and apply that!

lib/mediawiki.less
WalkQuackBack: These are translucent in the original, so I copied that. I think these should be a fade of accent, actually.

WalkQuackBack:
@uncenter Is this okay to merge given the approval, or is there any feedback I should act on?

lib/mediawiki.less
uncenter: Should this be changed first?

uncenter:
See above comment reply

WalkQuackBack:
Made the changes;

lib/mediawiki.less
```diff
@@ -66,79 +66,150 @@
...
+    --box-shadow-color-base: @overlay0;
+    --box-shadow-color-progressive--active: fade(@accent, 10%);
+    --box-shadow-color-progressive--focus: fade(@accent, 50%);
+    --box-shadow-color-progressive-selected: fade(@accent, 60%);
+    --box-shadow-color-progressive-selected--hover: fade(@accent, 70%);
+    --box-shadow-color-progressive-selected--active: fade(@accent, 80%);
+    --box-shadow-color-destructive--focus: fade(@accent, 50%);
+    --box-shadow-color-inverted: @base;
+    --box-shadow-color-alpha-base: rgba( 0, 0, 0, 0.87 );
+    --box-shadow-color-transparent: transparent;
```
uncenter, line 111: Do we need to set this?

lib/mediawiki.less
```diff
@@ -66,79 +66,150 @@
...
+    --background-color-error-subtle--active: lighten(mix(@red, @base, 30%), 10%);
+
+    --background-color-warning-subtle: mix(@yellow, @base, 30%);
+    --background-color-success-subtle: mix(@green, @base, 30%);
+    --background-color-notice-subtle: mix(@text, @base, 30%);
+
+    --background-color-content-added: mix(@blue, @base, 30%);
+    --background-color-content-removed: mix(@yellow, @base, 30%);
+
+    --background-color-transparent: transparent;
```
uncenter: Needs to be set?

lib/mediawiki.less
```diff
@@ -66,79 +66,150 @@
...
+    --background-color-backdrop-light: fade(@base, 65%);
+    --background-color-backdrop-dark: fade(@text, 65%);
+
+    --background-color-button-quiet--hover: fade(@accent, 2.7%);
+    --background-color-button-quiet--active: fade(@accent, 8.2%);
+
     --background-color-input-binary--checked: @accent;
+
+    --background-color-tab-list-item-framed--hover: rgba(255, 255, 255, 0.3);
+    --background-color-tab-list-item-framed--active: rgba(255, 255, 255, 0.65);
```
uncenter: Need to be set?

WalkQuackBack:
For all the set ones, those are set to have higher specificity over the existing theme on the website if they override those variables.

uncenter:
I'm just clarifying that by default, Mediawiki's value for `--background-color-transparent` isn't already `transparent`...?

WalkQuackBack:
> I'm just clarifying that by default, Mediawiki's value for `--background-color-transparent` isn't already `transparent`...?

I agree, the transparent variables don't need to be set again. Removing them.

lib/mediawiki.less
WalkQuackBack: Could this be set to a fade of Latte Base? It should always stay light, even in dark mode according to the style guide.

uncenter: Let's do that yeah.

WalkQuackBack:
```suggestion
    --background-color-tab-list-item-framed--hover: fade(@catppuccin[@latte][@base], 30%);
    --background-color-tab-list-item-framed--active: fade(@catppuccin[@latte][@base], 65%);
```

PR #1969, opened 2025-10-31
docs: update secondary action to contributing

uncenter:
We should link to our contributing documentation rather than directly to the GitHub I think. Icon shouldn't be external anymore, I'm thinking `rocket` but open to suggestions (https://starlight.astro.build/reference/icons/#all-icons).


sgoudham (dismissed):

uncenter:
image

This is what it looks like fwiw.

uncenter:
CleanShot 2025-11-03 at 18 32 41@2x

Changed the icons to match, I think it is better than with the rocket.

PR #1975, opened 2025-11-06
fix(graphite): domain from .dev to .com

cfbender:
## 🔧 What does this fix? 🔧
The main graphite domain has changed from `graphite.dev` to `graphite.com`. This updates it to match.

## 🗒 Checklist 🗒


uncenter (changes requested):
We should also change it at: https://github.com/catppuccin/userstyles/blob/7b21129d92ded3e4b9e027b6a9f7e689b96b1f3a/scripts/userstyles.yml#L352

PR #1977, opened 2025-11-08
docs(contributing/guides/raw-color-values): standard library links

ruiiiijiiiiang:
These two links in the contributing guides are currently broken.

image


uncenter:
The docs are (as of recently) meant to be viewed online at https://userstyles.catppuccin.com/contributing/ rather than through GitHub/locally. Those links are changed to work with Astro's requirements.

PR #1983, opened 2025-11-15
feat(proton): support lumo

ruiiiijiiiiang:
## 🔧 What does this fix? 🔧

Lumo is proton's latest AI chat bot service. It is mostly styled the same way as the other proton apps with some extra rules for generated code.

image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/proton/catppuccin.user.less
+++ b/styles/proton/catppuccin.user.less
@@ -243,6 +243,12 @@
 
     /* Proton Lumo */
 
+    header {
+      a[aria-label="Go to Lumo homepage"] img {
+        filter: @accent-filter;
+      }
+    }
+
     .whats-new-section {
       background: @accent;
     }


styles/proton/catppuccin.user.less
```diff
@@ -233,5 +238,66 @@
         );
       }
     }
+
```
uncenter, line 243: Probably worth adding a comment above here to note that the below styling is for the Lumo product specifically?

ruiiiijiiiiang: Comment added

uncenter:
This is a bit of a refactor but we shouldn't need to do `[@result]` - just `[]` should do the trick like with our other utilities. Worth a follow up PR.

uncenter:
Can we theme/filter the Lumo logo text?

ruiiiijiiiiang:
> Can we theme/filter the Lumo logo text?

image
It is now themed (I'm using lavender accent).

ruiiiijiiiiang:
> This is a bit of a refactor but we shouldn't need to do `[@result]` - just `[]` should do the trick like with our other utilities. Worth a follow up PR.

I can do the refactor either as part of this update or in a separate PR. Just let me know which you prefer.

uncenter:
> > Can we theme/filter the Lumo logo text?
> 
> image It is now themed (I'm using lavender accent).

Sorry, I meant this part:

CleanShot 2025-11-27 at 18 51 19@2x


> I can do the refactor either as part of this update or in a separate PR. Just let me know which you prefer.

Separate PR would be great after this one.

ruiiiijiiiiang:
> > > Can we theme/filter the Lumo logo text?
> > 
> > 
> > image It is now themed (I'm using lavender accent).
> 
> Sorry, I meant this part:
> 
> CleanShot 2025-11-27 at 18 51 19@2x
> > I can do the refactor either as part of this update or in a separate PR. Just let me know which you prefer.
> 
> Separate PR would be great after this one.

Done!
image

PR #1991, opened 2025-11-21
feat(arch-wiki): support chinese version

wz-willker:
(Apologies to Google Translate)
[Arch CN Wiki](https://wiki.archlinuxcn.org) is the Chinese version of Arch Wiki.
It uses the exact same style as Arch Wiki, so the same files can be reused.
I considered it part of Arch Wiki, so I didn't add an entry for it in userstyles.yml.

### DIFF
--- a/styles/arch-wiki/catppuccin.user.less
+++ b/styles/arch-wiki/catppuccin.user.less
@@ -17,7 +17,7 @@
 
 @import "https://userstyles.catppuccin.com/lib/lib.less";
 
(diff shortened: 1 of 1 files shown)


zenoix:
Can this be extended to the other arch wiki pages in other localisations like German and Japanese? If it can, might as well do it all at once?

hiruocha:
My test results show that it works for German, but not for Japanese.

PR #2008, opened 2025-11-30
refactor(proton): utility function no longer requires @result

ruiiiijiiiiang:
## 🔧 What does this fix? 🔧

Removed @result from utility function calls.

## 🗒 Checklist 🗒


uncenter:
This does still work right, functionally? Just confirming...
The trick is that `[]` is a catch-all return for all the values a mixin defines, which works when you only have one variable (as I understand it).

ruiiiijiiiiang:
> This does still work right, functionally? Just confirming... The trick is that `[]` is a catch-all return for all the values a mixin defines, which works when you only have one variable (as I understand it).

Yes, confirmed to still work.

uncenter (approved):

PR #2022, opened 2025-12-11
fix(lastfm): unreadable library bars text

LittleLilyBun:
On pages `/user/.../library/*`, the bars had unreadable text. I've fixed this by lowering the opacity of the bars to 20% (same as the date range bars on the same page) and changing the text colour from `crust` to `text`.
| Before | After |
| --- | --- |
|  |  |

This is my first time submitting a PR ever, so sorry if I made any mistakes!

### DIFF
--- a/styles/lastfm/catppuccin.user.less
+++ b/styles/lastfm/catppuccin.user.less
@@ -2,7 +2,7 @@
 @name Last.fm Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/lastfm
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/lastfm
-@version 2025.12.11
+@version 2025.12.27
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/lastfm/catppuccin.user.less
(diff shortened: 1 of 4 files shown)


styles/lastfm/catppuccin.user.less
```diff
@@ -2,7 +2,7 @@
 @name Last.fm Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/lastfm
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/lastfm
-@version 2025.09.06
+@version 2025.12.11
```
uncenter: It isn't necessary to update this version: https://userstyles.catppuccin.com/contributing/#versioning. You don't need to undo it, but just noting going forward :)

uncenter (approved):
This looks solid.

styles/lastfm/catppuccin.user.less
LittleLilyBun: Ah, sorry,. Must've missed that.

uncenter:
Note that I incorporated this change into #2037 already. @LittleLilyBun if you are interested in helping with a Lastfm rewrite, feel free to push to that branch. Otherwise you can keep pushing patches to the main branch for Lastfm :)

PR #1915, opened 2025-09-14
fix(reddit): background color behind subreddit name

kripikripi:
## 🔧 What does this fix? 🔧

Fixes background color behind subreddit name on posts

PR #2024, opened 2025-12-15
fix(reddit): theme credit bar background

tkimhong:
This fixes the` pdp-credit-bar` element's background color that shows the subreddit name, OP username, and time since it was posted (e.g. 3h ago) which was displaying with `--color-neutral-background` instead of matching the `@base` color.

The `--color-neutral-background` on both light/dark modes seems like Reddit's unthemed default colors.

## Before screenshots
**Latte:** 
image

**Macchiato:** 
image

## After screenshots
**Latte:** 
image

**Macchiato:**
image

### DIFF
--- a/styles/reddit/catppuccin.user.less
+++ b/styles/reddit/catppuccin.user.less
@@ -22,6 +22,7 @@
   :root:not(.theme-dark).theme-rpl,
   :root:not(.theme-dark) .theme-beta,
   :root:not(.theme-dark).theme-beta,
+  :root:not(.theme-dark) #pdp-credit-bar,
   .theme-light {
     #catppuccin(@lightFlavor);
   }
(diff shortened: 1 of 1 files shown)


uncenter:
Duplicate of https://github.com/catppuccin/userstyles/pull/1915 but we can stick with this new active one I guess. The one concern is that this isn't the ideal way to fix the problem. We may want to change where the variables are applied instead, as according to the original CSS there are a few higher specificity components to override for variables.

tkimhong:
> Duplicate of #1915 but we can stick with this new active one I guess. The one concern is that this isn't the ideal way to fix the problem. We may want to change where the variables are applied instead, as according to the original CSS there are a few higher specificity components to override for variables.

Sorry, I didn't see that. This is my first time contributing to open source, so apologies if I made mistakes.

After your feedback, I've done some digging in Reddit's styles pattern and refactored the fix to add `:root:not(.theme-dark) #pdp-credit-bar` to the high specificity selector groups (line 25 and 34) so it applies the whole theme. What do you think of this approach?

PR #2030, opened 2025-12-23
fix(go.dev): CVE banner

givensuman:
## 🔧 What does this fix? 🔧

The CVE notice at the top of some Go package pages was unstyled and near unreadable.

Screenshot_2025-12-23_17-46-12

There's no good way to do this with a red background, so I opted for a border instead.

Screenshot_2025-12-23_17-47-15

## 🗒 Checklist 🗒

### DIFF
--- a/styles/go.dev/catppuccin.user.less
+++ b/styles/go.dev/catppuccin.user.less
@@ -2,7 +2,7 @@
 @name go.dev Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/go.dev
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/go.dev
-@version 2025.09.06
+@version 2025.12.23
 @updateURL https://github.com/catppuccin/userstyles/raw/main/styles/go.dev/catppuccin.user.less
 @supportURL https://github.com/catppuccin/userstyles/issues?q=is%3Aopen+is%3Aissue+label%3Ago.dev
 @description Soothing pastel theme for go.dev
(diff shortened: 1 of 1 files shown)


givensuman:
Hmmm the linter doesn't seem to be satisfiable.

```
border-left-width: 4px;
border-left-style: solid;
border-left-color: @red;
```
gives the error `Expected shorthand property "border-left"`, as written it gives the error `Use "border-left-color" instead of "border-left"`

styles/go.dev/catppuccin.user.less
```diff
@@ -2,7 +2,7 @@
 @name go.dev Catppuccin
 @namespace github.com/catppuccin/userstyles/styles/go.dev
 @homepageURL https://github.com/catppuccin/userstyles/tree/main/styles/go.dev
-@version 2025.09.06
+@version 2025.12.23
```
uncenter, line 5: Doesn't need to be manually edited, see https://userstyles.catppuccin.com/contributing/#versioning. No need to revert now but please read our contributing documentation going forward!

styles/go.dev/catppuccin.user.less
```diff
@@ -433,6 +433,11 @@
       content: url("data:image/svg+xml,@{svg}");
       filter: none;
     }
+    // CVE notice
+    .go-Message--alert {
+      background-color: @base;
+      border-left: solid 4px @red;
+    }
```
uncenter, line 440: If the background is red you can set the text to `crust` for contrast and that should do the trick :)

PR #2037, opened 2025-12-27
feat(lastfm): rewrite

uncenter:
Rewrites the Lastfm userstyle, ~1800 to ~800 lines. Most areas are themed, still some work left to do though. I think this is more maintainable and accurate/up to date.

### DIFF
--- a/.github/ISSUE_TEMPLATE/userstyle.yml
+++ b/.github/ISSUE_TEMPLATE/userstyle.yml
@@ -7,7 +7,7 @@ body:
       label: Is there an existing issue outlining your problem?
       description: Please search to see if an issue already exists for your problem.
       options:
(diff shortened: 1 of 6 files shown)


AnubisNekhet (commented):
Tested it out, LGTM

uncenter:
Fwiw the two todo pages are https://www.last.fm/music and https://www.last.fm/user/*/listening-report/week (Reports). The later is all unthemed, which is good in that it isn't half themed lol. Could theme it in one pass at some point, the old theming of it wasn't fully working either so I think its fine to leave it out here.

entraptaa:
Is this why the song covers are all inverted? 
image

uncenter:
> Is this why the song covers are all inverted?

Oops, that's a bug! In the future please open new issues for those. Fixed in #2049.

PR #1870, opened 2025-09-01
feat(arch-user-repository): init

opensourcecheemsburgers:
## 💬 Additional Comments 💬

The main difficulties I had with this port were during the initial setup.

Issues:
- Stylus did not react to changes on my `catppuccin.user.less`, so I made a symlink to `catppuccin.user.css`.
- Understanding that I should not install the style, but only enable 'Live Reload'.
- LESS is currently [not supported](https://github.com/helix-editor/helix/pull/14303) in Helix.
- Firefox does not allow you to set the color of the `select`'s `option` children. It is based off of the current theme.

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [x] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [x] I have ensured that the new directory is in **lower-kebab-case**.
  - [x] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have made sure to update the [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml) file with information about the new userstyle.
- [x] I have included the following files:
  - [x] `catppuccin.user.less` - all the CSS for the userstyle, based on the template.

PR #1381, opened 2024-10-16
feat/arch-user-repository

claymorwan:
## 🎉 Theme for Arch user repository 🎉

The Arch user repository (aur) is a community-driven repository of packages for Arch Linux.
image

## 💬 Additional Comments 💬

I had a bit of struggle making this theme since it was my first time working with userstyles, but I made it in a day so it's fine. My only problem at the moment is changing the logo for a themed one, or at least a transparent one.

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://github.com/catppuccin/userstyles/blob/main/docs/userstyle-creation.md).
- [x] I have made a new directory underneath `/styles/<name-of-website>` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.
  - [x] I have ensured that the new directory is in **lower-kebab-case**.
  - [x] I have followed the template and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have made sure to update the
      [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml)
      file with information about the new userstyle.
- [x] I have included the following files:
  - [x] `catppuccin.user.css` - all the CSS for the userstyle, based on the
        template.
  - [x] `preview.webp` - composite image of all four individual flavor screenshots (taken with the default accent color of mauve) stitched together, generated via [Catwalk](https://github.com/catppuccin/catwalk).

PR #1895, opened 2025-09-08
refactor(wiki.nixos.org): use mediawiki library module

uncenter:
Integrates the new library module for MediaWiki sites from #1866. **Must be merged after #1866 is merged into main.**

Closes #1864

PR #2043, opened 2025-12-28
feat(arch-linux): init

ninetailedtori:
Closes #1870, closes #1381.

## 🎉 Theme for Arch Linux 🎉

Honestly, this is a major refactor of the existing arch-wiki theme into an umbrella "Arch Linux" userstyle, including:
- arch-wiki existing theme
  - [x] migrated to mediawiki variables?
- an additional one started by someone else for aur.archlinux.org: #1870 
  - [x] merged with this PR yet?
  - [x] finished?
- and all the others I'll be actively doing for Arch Linux!
  - [x] archlinux (home)?
  - [x] archlinux/download?
  - [x] man.archlinux?
  - [x] archlinux/packages?
  - [x] security.archlinux?
  - [x] bbs.archlinux?

## 💬 Additional Comments 💬

This is based off of the old PR that died, #1381, but will be moving the wiki to use MediaWiki variables in the library module e.g. #1895, once we can confirm it working as it should on arch's wiki (no leftover overriding on their end). All other sites will be done by me, aside from the AUR one which was done by someone else (though it'll be edited by me as well, to fall in line with the new style of formatting).

**Base: #1866**

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://userstyles.catppuccin.com/contributing/creating-userstyles/).
- [x] I have made a new directory underneath `/styles/archlinux` containing the contents of the [`/template`](https://github.com/catppuccin/userstyles/blob/main/template/) directory.


styles/archlinux/README.md
uncenter, line 1: No need to rename, just delete the file and it will be re-generated properly once merged :)

styles/archlinux/catppuccin.user.less
```diff
@@ -0,0 +1,386 @@
...
+
+    ::selection {
+      background-color: fade(@accent, 30%) !important;
+    }
+
+    input,
+    textarea {
+      &::placeholder {
+        color: @subtext0 !important;
+      }
```
uncenter: We don't need this because it comes with `#lib.defaults();`. The input/textarea background color setting should be separated out.

styles/archlinux/catppuccin.user.less
```diff
@@ -0,0 +1,386 @@
...
+        );
```
uncenter: This logo appears to be the exact same as on line 136 for the manpages. We can probably consolidate these? Perhaps a base `@-moz-document domain("archlinux.org")` that applies to all the subdomain pages?

ninetailedtori: That's actually what I was going to do, pushing that change right now xD

styles/archlinux/catppuccin.user.less
ninetailedtori: Thanks for catching that one! I think it should be cleaned up in the next commit o7

ninetailedtori:
PR is now up to every page aside from:
- ~~Download~~ DONE!
- Wiki (needs mediawiki lib, will set up my own localhost tomorrow)
Feel free to test it for me! Security and BBS were the last ones I just did, and rework of AUR as well as expansion of it to include Packages (official) was done earlier as well! :]

styles/archlinux/catppuccin.user.less
```diff
@@ -246,9 +246,31 @@
...
-    select option,
+    input {
+      color: @text !important;
+      background-color: @surface0 !important;
+      border-color: @surface1 !important;
+
+      &:hover {
+        background-color: @surface1 !important;
+        border-color: @surface2 !important;
+        cursor: pointer !important;
```
uncenter: We shouldn't set this.

styles/archlinux/catppuccin.user.less
```diff
@@ -246,9 +246,31 @@
...
+
+      &:active, &:enabled {
+        background-color: fade(@overlay2, 30%) !important;
+        border-color: @overlay0 !important;
+      }
+    }
+
+    input[type="reset"] {
+      background-color: @red !important;
+      padding: 0 2px !important;
```
uncenter: Is this necessary?

styles/archlinux/catppuccin.user.less
```diff
@@ -246,9 +246,31 @@
...
+      border-color: @maroon !important;
+      color: @base !important;
+      cursor: pointer !important;
```
uncenter: Ditto.

uncenter (commented):
Looks like there is still a bunch of duplicated input styles?

styles/archlinux/catppuccin.user.less
ninetailedtori: Why not? You mean the cursor? That lingered from the original AUR pr actually, should I remove it?

styles/archlinux/catppuccin.user.less
ninetailedtori: The padding? I'll remove and see. Another from og AUR pr, good check o7

styles/archlinux/catppuccin.user.less
uncenter: Yes please! Our policy is to not change non-color related styling properties.

styles/archlinux/catppuccin.user.less
ninetailedtori: Yes, this is necessary so the button doesn't look cramped, unfortunately. Is it ok to keep?

uncenter: The fixing the cramped looking button is an aesthetic choice, change it upstream rather than here :)

ninetailedtori:
> Looks like there is still a bunch of duplicated input styles?

If you spot any duplicated styles that aren't being annoyingly overwritten by *something* on the page (like the .pun #brdX regions), do lemme know o7
I'm not sure how to get them to overwrite including on top of the other css used on the page without specificity unfortunately.

styles/archlinux/catppuccin.user.less
ninetailedtori: Removed, will open a PR upstream (once I can work out how they've bloody packaged their site o-o)

ninetailedtori:
This is a reminder for me to change the mediawiki location once that's published, because I'm *going* to forget :P

WalkQuackBack:
This may have rebased this onto the wrong branch. Could you try rebasing onto origin/main again?

ninetailedtori:
Shoot, tried to merge but not rebase into mediawiki ;-;

scripts/userstyles.yml
```diff
@@ -109,14 +111,13 @@ userstyles:
         name: AniChart
         link: https://anichart.net
     current-maintainers: [*AnubisNekhet]
-  arch-wiki:
-    name: Arch Wiki
-    link: https://wiki.archlinux.org
-    categories: [wiki]
+  archlinux:
```
uncenter: I would change this to `arch-linux`.

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1269 @@
...
+@preprocessor less
+@var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]
+@var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
+@var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+==/UserStyle== */
+
+@import "https://userstyles.catppuccin.com/lib/lib.less";
+@import "http://127.0.0.1:8000/lib/mediawiki.less";
+
+/* deno-fmt-ignore */
```
uncenter: Why are we disabling formatting?

styles/archlinux/catppuccin.user.less
```diff
@@ -0,0 +1,1269 @@
...
+  }
+
+  #catppuccin(@flavor) {
+    #lib.palette();
+    #lib.defaults();
+    #lib.css-variables();
+
+    --bg-color: @base  !important;
+    --color: @text  !important;
+    --code-inline-bg-color: @mantle  !important;
```
uncenter: Probably undocumented but I think we usually do surface0 for inline code blocks.

```suggestion
    --code-inline-bg-color: @surface0  !important;
```

styles/archlinux/catppuccin.user.less
```diff
@@ -0,0 +1,1269 @@
...
+        );
+        background: url("data:image/svg+xml,@{svg}")  !important;
+        background-size: 190px 40px  !important;
```
uncenter: Do we need to set this?

uncenter (commented):
There's a lot of duplication between the first @-moz-document and for bbs.archlinux.com. Can we avoid that? Does the regex for the first one not include bbs?

ninetailedtori:
> There's a lot of duplication between the first @-moz-document and for bbs.archlinux.com. Can we avoid that? Does the regex for the first one not include bbs?

The regex DOES apply to bbs, it just isn't important enough for some reason.

I tried, but where I redefined, it wasn't applying at all. I might actually define all of these commons in a modular way, then just call the function within the required scopes at this point. Not sure why they had something overwrite it though, very annoying. I'll see what I can do though, through specificity tinkering.

styles/arch-linux/catppuccin.user.less
ninetailedtori: It splits the regexp onto multiple lines, which unfortunately breaks stylus extension rn.

styles/archlinux/catppuccin.user.less
ninetailedtori: Errrr, where's that from? Logo? Yes, we have to set that. Because we're replacing the old logo, with catppuccin coloured one, and without setting that it breaks. Source: tested it :]
image

ninetailedtori: image

This is what current code looks like, i.e. size is fixed by our setting manually.

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1292 @@
...
+@-moz-document regexp("^https:\\/\\/(?:[-a-zA-Z0-9@:%_+.~#?&\\/=]*\\.)?(?<!gitlab\\.)archlinux\\.org\\/?[-a-zA-Z0-9@:%_+.~#?&\\/=]*") {
```
uncenter:
```suggestion
@-moz-document regexp("^https:\/\/(?!gitlab\.)(.+\.)*archlinux\.org.*") {
```

I feel like this simplified version does the same thing?

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1292 @@
...
+    #lib.css-variables();
+
+    #archnavbar ul li#anb-aur a {
+      color: @accent !important;
+    }
+  }
+}
+
+/* deno-fmt-ignore */
+@-moz-document regexp("^https:\\/\\/archlinux\\.org\\/?[-a-zA-Z0-9@:%_+.~#?&\\/=]*"), domain("aur.archlinux.org") {
```
uncenter: What's this first regex supposed to be? Seems like it could instead be `url-prefix("https://archlinux.org")`?

uncenter:
> I might actually define all of these commons in a modular way, then just call the function within the required scopes at this point.

This is what I would recommend doing.

> Not sure why they had something overwrite it though, very annoying. I'll see what I can do though, through specificity tinkering.

You could do a `&, .pun` wrapper for everything in the first moz-document and that would do the trick I think, a little ugly though.

styles/arch-linux/catppuccin.user.less
ninetailedtori: You need double backslash in stylus usercss regex, actually!

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1347 @@
...
+    #lib.css-variables();
+
+    #archnavbar ul li#anb-aur a {
+      color: @accent !important;
+    }
+  }
+}
+
+/* deno-fmt-ignore */
+@-moz-document regexp("^https:\\/\\/(?!gitlab\\.)(\\w+\\.)*archlinux\\.org\\/(?:packages[\\/]?)\\w*"), domain("aur.archlinux.org") {
```
uncenter: Are there any subdomains of packages or does this work the same?
```suggestion
@-moz-document url-prefix("https://archlinux.org/packages"), domain("aur.archlinux.org") {
```

ninetailedtori: Oh shoot- wait; you're completely right, yea lemme do that.

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1347 @@
...
+    #lib.css-variables();
+
+    #archnavbar ul li#anb-home a {
+      color: @accent !important;
+    }
+  }
+}
+
+/* deno-fmt-ignore */
+@-moz-document regexp("^https:\\/\\/(?!gitlab\\.)(\\w+\\.)*archlinux\\.org\\/(?:packages[\\/]?)\\w*") {
```
uncenter:
```suggestion
@-moz-document url-prefix("https://archlinux.org/packages") {
```

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1347 @@
...
+        background: url("data:image/svg+xml,@{svg}") !important;
+        background-size: 190px 40px !important;
```
uncenter:
```suggestion
        background-image: url("data:image/svg+xml,@{svg}") !important;
```

ninetailedtori: Not possible to remove size, it breaks the image.

uncenter: Yes it is, if you set the `background-image` instead of overriding `background`. Try it yourself.

ninetailedtori:
> > I might actually define all of these commons in a modular way, then just call the function within the required scopes at this point.
> 
> 
> 
> This is what I would recommend doing.
> 
> 
> 
> > Not sure why they had something overwrite it though, very annoying. I'll see what I can do though, through specificity tinkering.
> 
> 
> 
> You could do a `&, .pun` wrapper for everything in the first moz-document and that would do the trick I think, a little ugly though.

I'm going to do the wrapper, tragically there's no solution. I tried refactoring to remove dupes outright and it broke as like last time I tested. I'll wrap the whole thing I think.

styles/arch-linux/catppuccin.user.less
ninetailedtori: Oh shoot, really? Wait, lemme try that!

styles/arch-linux/catppuccin.user.less
ninetailedtori: I think so? Might be wrong, lemme check. I'll add comments in the next commit!

uncenter:
> I'm going to do the wrapper, tragically there's no solution. I tried refactoring to remove dupes outright and it broke as like last time I tested. I'll wrap the whole thing I think.

Did you try making mixins and then just applying those in the right places? I see no reason why that wouldn't work if done correctly.

styles/arch-linux/catppuccin.user.less
uncenter: Yep I copied from my regex tester instead of my source file, sorry.

uncenter:
Generally we should avoid regex if possible. Hard to read, hard to maintain, often there are better solutions :)

uncenter:
Fwiw here is a good playground that I like to use: https://regex101.com/r/wrKYCM/1

ninetailedtori:
> Fwiw here is a good playground that I like to use: https://regex101.com/r/wrKYCM/1

That's what I've been using, have been for years! xD
Unfortunately it seems that `\w` is broken in stylus, which means we DO have to use `.`

ninetailedtori:
> Generally we should avoid regex if possible. Hard to read, hard to maintain, often there are better solutions :)

Also slower, so you're completely right on that. It's just in my regex mania I've completely forgotten to see if I could make it easier with just prefixing xD

ninetailedtori:
> > I'm going to do the wrapper, tragically there's no solution. I tried refactoring to remove dupes outright and it broke as like last time I tested. I'll wrap the whole thing I think.
> 
> Did you try making mixins and then just applying those in the right places? I see no reason why that wouldn't work if done correctly.

Actually, this was my entire plan xD

uncenter:
> Unfortunately it seems that `\w` is broken in stylus, which means we DO have to use `.`

Yeah there isn't any reason we need to use \w over `.`. We aren't parsing or validating much here.

uncenter:
> That's what I've been using, have been for years! xD

Just saying I put the test strings for each of the URLs in there so it's easier.

ninetailedtori:
> > That's what I've been using, have been for years! xD
> 
> Just saying I put the test strings for each of the URLs in there so it's easier.

Ya, that's what I've been doing! :3 it is annoying though, I have to then test it twice, because stylus regex isn't fully with the same accepted tokens as it is in that tester, so I have to check every time what sites are recognised. xD anyway rn I'm doing mixins, so will be big update soon!

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1501 @@
...
+        }
+      }
+
+      span.arrow {
+        color: transparent !important;
+        background: none !important;
+        border-color: @blue transparent transparent transparent !important;
+      }
+    }
+  }
```
uncenter: I don't understand the styling here. This doesn't match the original site.

Image

Image

This could just be this, according to the styling of the original site:

```css
      a, a:link, a:visted {
        color: @accent !important;
      }
```

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1501 @@
...
+          background-color: @surface2 !important;
+        }
+      }
+
+      td {
+        border-color: @overlay0 !important;
+      }
+
+      th {
+        color: @peach !important;
```
uncenter: Why are we making these peach if the original are just text colored?

Image

Image

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1501 @@
...
+    }
+
+    #footer, .footer {
+      color: @subtext0 !important;
+    }
+
+    // FIXME: This doesn't work yet, but I want to just apply filter to all span icons :]
+    span[background-image="url(Air/img/bull.png)"] {
+      filter: @accent-filter !important;
+    }
```
uncenter: This seems over complicated. The base styles should just be:

```suggestion
    body {
      color: @text;
      background-color: @base;
    }
```

The rest of these are unnecessary/should go somewhere else.

ninetailedtori: global labels are required, because otherwise we'd have to add about 30 different places to set it. It's equivalent for all of them unless otherwise overwritten, which takes a much lower line of code. Footers are the same, every page has a footer. Legends the same, there's a lot of legends, similar to labels, used just as much. Content I can see if I can simplify, but I did try iirc, and there were a lot of things that broke.

styles/arch-linux/catppuccin.user.less
ninetailedtori: Headings, ignore it, it's getting redone.

styles/arch-linux/catppuccin.user.less
uncenter: Let's at the least avoid `content *`.

ninetailedtori: Done, safely removed.

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1602 @@
...
+@license MIT
+
+@preprocessor less
+@var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]
+@var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
+@var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+==/UserStyle== */
+
+@import "https://userstyles.catppuccin.com/lib/lib.less";
+@import "http://127.0.0.1:8000/lib/mediawiki.less";
```
WalkQuackBack: Reminder to change this before merging.

styles/arch-wiki/catppuccin.user.less
WalkQuackBack, line 1: We should keep this around for at least a few months, with the updateURL pointing to the new userstyle, in a similar way to the CSS > Less migration. Then, current users will be able to migrate properly.

uncenter: @ninetailedtori we will take care of this, don't worry about doing this yourself.

styles/arch-linux/catppuccin.user.less
```diff
@@ -0,0 +1,1931 @@
...
+@license MIT
+
+@preprocessor less
+@var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]
+@var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
+@var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+==/UserStyle== */
+
+@import "https://userstyles.catppuccin.com/lib/lib.less";
+@import "http://127.0.0.1:8000/catppuccin-styles/userstyles/lib/mediawiki.less";
```
ninetailedtori, line 19:
```suggestion
@import "https://userstyles.catppuccin.com/lib/mediawiki.less";
```

PR #2005, opened 2025-11-28
docs: add userstyles review tips page

uncenter:
- [ ] Links
- [ ] System elements

PR #2046, opened 2025-12-29
docs: add userstyles style guide

uncenter:


uncenter:
- [ ] Code block backgrounds (inline and block)
	- `surface0` for inline, `mantle` for block
- [ ] Tooltip backgrounds
	- `mantle`/`crust` (darker) vs `surface0` (lighter)
- [ ] Link colors
	- [ ] Exception for userstyles with interfaces where the "primary interactive elements" are links (e.g. search engines). ref https://github.com/catppuccin/userstyles/pull/1450#pullrequestreview-3633929541

uncenter:
Todo:

> 1. How to handle shadows (e.g. text shadows, box shadows). This seems to be inconsistent across the board.
> 2. When is it appropriate to use !important vs having a more specific selector for code clarity?
> 3. Hover, active and focus states. How much should they be lightened or darkened by? Should these be standardized into the library? In theory, tints and shades should help with this, but a solution should be had in the meantime.
> 4. Faded backgrounds such as diff code added/removed. Should the fade or mix amount on this be standardized?
> 5. Application of mixin: way it should be applied. As an example, some userstyles do `body:not(.dark)` and `body.dark` or `[data-theme=“light”]` and `[data-theme=“dark”]` while others do `body` and `body.dark` or `mroot` and `&[data-theme=“dark”]`.
> 6. Encourage using mixins for repetitive code
> 7. Specify how apps with colour ramps should be bound to Catppuccin. Atp, this is all over the place.
> 8. How should semi-transparent backgrounds be handled?
> 9. Userstyles should prefer using attribute selectors over generated classes. I am personally not a fan of using localization specific selectors though.
> 10. Usage of vendor prefixes and experimental/not baseline features should be mentioned. 

(https://github.com/catppuccin/userstyles/pull/2005#issuecomment-3591969266)

PR #2053, opened 2026-01-03
docs: mark flaky userstyles

uncenter:


uncenter:
Curious what people think of this. We mark "badly broken" or instable userstyles as flaky in the metadata, which has the result of a new emoji code on the listing and disables them by default in the all userstyles import. Pull request needs to be updated but the idea is the same.

PR #2056, opened 2026-01-04
feat(lib/mediawiki): add a.extiw

ninetailedtori:
### Current Suggestion:

- Adds `a.extiw` - interwiki links icon!
- Adds check for `:not(.plainlinks)` - plainlinks have no mw parsing, and thus cause the image to be repeated at every single letter's x position. Add a check, to only apply our icon to all non-plainlinks.

## 🗒 Checklist 🗒

### DIFF
--- a/lib/mediawiki.less
+++ b/lib/mediawiki.less
@@ -253,7 +253,7 @@
       }
     }
 
-    .mw-parser-output a {
+    :where(.mw-parser-output a) {
       &.external {
         @svg: escape(
           '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" viewBox="0 0 12 12"><path fill="@{accent}" d="M6 1h5v5L8.86 3.85 4.7 8 4 7.3l4.15-4.16zM2 3h2v1H2v6h6V8h1v2a1 1 0 0 1-1 1H2a1 1 0 0 1-1-1V4a1 1 0 0 1 1-1"/></svg>'


lib/mediawiki.less
```diff
@@ -238,8 +243,24 @@
...
+  }
+
+  .icons {
+    .mw-parser-output {
+      a.external {
+        @svg: escape(
+          '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" viewBox="0 0 12 12"><path fill="@{accent}" d="M6 1h5v5L8.86 3.85 4.7 8 4 7.3l4.15-4.16zM2 3h2v1H2v6h6V8h1v2a1 1 0 0 1-1 1H2a1 1 0 0 1-1-1V4a1 1 0 0 1 1-1"/></svg>'
+        );
+        background-image: url("data:image/svg+xml,@{svg}");
+      }
```
uncenter: As noted at https://github.com/catppuccin/userstyles/pull/1866#discussion_r2658970220, we should leave this as is. The issue with the icon appearing in the wrong place(s) was that it wasn't respecting the `.plainlinks` class. As such, we should respect it by using `:not(.plainlinks) a.external` as the selector rather than moving it just for !important.

ninetailedtori: Should we do that for all cases, wiki specific? Archwiki with plainlinks, and Wikipedia with ext-interwiki? But do we actually want to provide the icons for those or simply not? Would that not be inconsistent?

uncenter: Sorry, "Wikipedia with ext-interwiki"? I found the `.plainlinks` class originally on Wikipedia, for starters. I haven't seen ext-interwiki yet.

ninetailedtori: > Sorry, "Wikipedia with ext-interwiki"? I found the `.plainlinks` class originally on Wikipedia, for starters. I haven't seen ext-interwiki yet.

"extiw", they're interwiki links, but this is due to me adding the icon for interwiki links for wikipedia. it's probably best to have an actual check to see if a background-image has been defined, then apply our image settings.

ninetailedtori: See: https://en.wikipedia.org/wiki/Help:Interwiki_linking

uncenter: Where on that page can I see the icon for interwiki links?

ninetailedtori: You can't, but the class exists and is used by Wikipedia a lot. To see interwiki icons, see other wiki sites like arch wiki, where the only icons defined are interwiki icons. It's the reasoning behind why interwiki icons are intended to be defined, as a base icon from wikimedia.

uncenter: Can you link to a page where the interwiki icon is used?

> You can't, but the class exists and is used by Wikipedia a lot.

So they don't enable the visual icon if I understand correctly?

lib/mediawiki.less
```diff
@@ -208,7 +209,36 @@
...
+    --border-color-success: @green;
+    --border-color-notice: @surface2;
+    --border-color-content-added: @blue;
+    --border-color-content-removed: @peach;
+    --border-color-progressive: @accent;
+    --border-color-progressive--active: @accent;
+    --border-color-progressive--hover: @accent;
+    --border-color-progressive--focus: @accent;
+  }
+>>>>>>> 5c2736de (Attempt at modularising it?)
```
uncenter: Looks like we have some merge artifacts :)

lib/mediawiki.less
```diff
@@ -266,6 +287,18 @@
...
+    .vector-toc-pinned-container {
+      background-color: inherit;
+      border: none;
+    }
+
+    /* All other pinned containers */
+    .vector-pinned-container:not(.vector-toc-pinned-container) {
+      background-color: @mantle;
+      border-color: @overlay0;
+    }
```
uncenter: Reproduction URL? Is this still necessary with our updated Codex variables? I'd like to see if there is a cleaner way of doing this.

ninetailedtori: Try this one: https://wiki.archlinux.org/title/Arch_Linux!
Some other wikis that use it are: https://www.mediawiki.org/wiki/Skin:Vector/2022
And I think wikipedia uses that now as well? The idea is to make the Table of Contents merge cleanly with the bg, like wikis normally are, which was my thinking, and other pinned containers to appear in hover. I *think* there's another url for pinned non toc containers but I can't find where it was, but I definitely had one somewhere.

ninetailedtori:
@WalkQuackBack we should prolly add a bug tag to this, now it's a sole fix :3

ninetailedtori:
Not certain this one is needed. Best to perhaps move it into my own lib, and add a snippet to the mediawiki usage docs, if needed, but not to mess up the module as it breaks on pages that don't set an image for extiw.

PR #2060, opened 2026-01-04
refactor(google): auto build script

WalkQuackBack:


styles/google/scripts/polyfills/legacy.css
uncenter, line 1: What is this from? Is this the output?

WalkQuackBack: This is a polyfill written by me to handle the legacy components like the ones on [Google Advanced Search](https://www.google.com/advanced_search) that don't bind well to Material 3 with the gradients and such.

PR #2070, opened 2026-01-06
fix(status.cafe): style guide links

uncenter:
Part of an effort to standardize on our style guide - as per the style guide, links should be blue and visited ("followed") links should be lavender.


uncenter:
Hold off on merging for now actually.

bunfluff:
i am okay with this being merged but i am a bit confused: what would be done with the accent colors? due to status.cafe's simple nature, the only semblance to accents that could be applied would be on links

PR #2088, opened 2026-01-16
feat(git-scm): init

ninetailedtori:
## 💬 Additional Comments 💬

Draftie-draft! I'll be documenting this as I go along! It's not even in a reviewable state now! :3

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [submission guidelines](https://userstyles.catppuccin.com/contributing/creating-userstyles/).


ninetailedtori:
if only I could pause lint till this was non-draft...unless

uncenter:
> if only I could pause lint till this was non-draft...unless

Just ignore it.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+
+        &:hover {
+          background-color: var(--pre-bg-color) !important;
+        }
+      }
+    }
+
+    input.inactive {
+      color: var(--light-font-color) !important;
+    }
```
uncenter, line 427: I can't tell if these are our overrides or theirs. If they are ours, we should use our Less variables.

ninetailedtori: They're our overrides, but I wanted to tie everything back to their variables, so that I could cleanly map their variables to our less variables, if that makes sense! It's easier to control them all at once then.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+
+    --orange: @accent;
+    --orange-darker-5: darken(@accent, 5%);
+    --blue: @blue;
+    --font-color: @text;
+    --aside-font-color: @subtext0;
+    --light-font-color: @subtext0;
+    --light-font-color-darker-10: @text;
+    --light-font-color-darker-25: @base;
+    --light-font-color-darker-35: @text;
```
uncenter: Would be good to leave comments here indicating where these are used, otherwise it's confusing why darker-25 is base in a sea of text.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+    --light-font-color-darker-55: darken(@subtext0, 55%);
+    --light-font-color-lighter-20: lighten(@subtext0, 20%);
```
uncenter: I think I mentioned this elsewhere but debatable if we should follow their exact values and exactly darken/lighten to their colors. Remember, we want to make it look properly Catppuccin, and avoid modifying colors from the palette too much when possible.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+      darken(@accent, 5%),
+      darken(@accent, 10%)
+    );
+    --button-color: @text;
+    --button-border: @overlay0;
+    --mark-bg-color: fade(@yellow, 30%);
+    --mark-color: @yellow;
+    --base-border-color: @overlay0;
+    --base-border-color-darker-8: darken(@overlay0, 8%);
+    --base-border-color-lighter-10: darken(@overlay0, 10%);
```
uncenter: Ditto from earlier, use shades here ideally because we do have the surfaces and the overlays to use.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+    --callout-color: @mantle;
+    --highlight-bg-color: fade(@accent, 30%);
+    --pre-bg-color: @mantle;
+    --pre-border: @overlay0;
+    --synopsis-bg-color: @surface0;
+    --black-3: @crust;
```
uncenter, line 81: Some of these more ambiguous variables should also have comments explaining where/why.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+    --logos-4: @accent;
+    --logos-5: darken(@accent, 25%);
+    --logos-6: darken(@accent, 30%);
+    --logos-7: lighten(@accent, 5%);
+
+    img[alt="Git"] {
+      filter: @accent-filter !important;
+    }
+
+    // Thank gosh, the icons need to be the same color as the text anyway!
```
uncenter: A more helpful comment would be nice :)
```suggestion
    // Icons next to text headings are added using a background on the h3 element; filtering the element themes the icon and has no visible changes to text (same color used already).
```

Ideally we would use a subtext0 or something for the icon since it is fainter in the original but this will do for the reasons provided.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+
+    button {
+      border-color: var(--button-border) !important;
+    }
+
+    fieldset {
+      border-color: var(--pre-border) !important;
+    }
+
+    // ! Not sure we're keeping this. Keep an eye on Leah's decision.
```
uncenter: Can you remind me what this is about?

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+    input.inactive {
+      color: var(--light-font-color) !important;
+    }
+
+    form#search {
+      @svg: escape(
+          '<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16"><path fill="none" stroke="@{subtext0}" stroke-linecap="round" stroke-linejoin="round" d="M12.5 6.5a6 6 0 01-6 6 6 6 0 01-6-6 6 6 0 016-6 6 6 0 016 6m3 9L11 11"/></svg>'
+        )
+        !important;
+      color: var(--aside-font-color) !important;
```
uncenter: This doesn't seem like the appropriate variable here.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+        color: var(--font-color) !important;
+        background-image: url("data:image/svg+xml,@{svg}") !important;
+      }
+
+      box-shadow: var(--shadow-color) !important;
+    }
+
+    #documentation #main {
+      p.nutshell, .literalblock .content pre tt {
+        background-color: var(--pre-bg-color) !important;
```
uncenter: Pre should be for pre, not anything else. Probably should use a more appropriate border color.

styles/git-scm/catppuccin.user.less
uncenter: Tentatively we can let it go forward like this as we discuss, so long as the mappings use appropriate variables as to avoid confusion later.

styles/git-scm/catppuccin.user.less
```diff
@@ -0,0 +1,622 @@
...
+@license MIT
+
+@preprocessor less
+@var select lightFlavor "Light Flavor" ["latte:Latte*", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha"]
+@var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
+@var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
+==/UserStyle== */
+
+@import "https://userstyles.catppuccin.com/lib/lib.less";
+@import url("https://python.catppuccin.com/pygments/catppuccin-variables.important.css");
```
uncenter: Please follow our instructions about precisely where highlighting import URLs should be placed: https://userstyles.catppuccin.com/contributing/guides/syntax-highlighting/. Right now, with this userstyle installed, Stylus considers it applied to every single webpage because this import adds CSS that isn't scoped via `@-moz-document`. It should be moved to between lines 21 and 22.

styles/git-scm/catppuccin.user.less
ninetailedtori: That sounds good! :]

styles/git-scm/catppuccin.user.less
ninetailedtori: That's true! Let me see if I can just change it to the actual colours on the palette without much issue.

styles/git-scm/catppuccin.user.less
ninetailedtori: Resolved in next commit!

styles/git-scm/catppuccin.user.less
ninetailedtori: those are pres, actually. That's how they're referred to in the git-scm reference manual. Even upstream, we see p.synopsis (which is the header block that lists all command options, similar to manpages), p.nutshell (used in the book reference) and the .literalblocks are actually codeblocks but for holding codeblocks of file contents or stdouts etc., etc referred to with pre-border, so it's just an override to make them more sensible.

styles/git-scm/catppuccin.user.less
ninetailedtori: Whether we want to overwrite the inline code formatting so it's like our inline code setting.

styles/git-scm/catppuccin.user.less
ninetailedtori: Fixed in next commit!

styles/git-scm/catppuccin.user.less
uncenter: Please note that in the comment otherwise this is ambiguous :)

ninetailedtori:
> Do we need !important on everything?

Uncertain, some stuff WAS necessary but I can't remember how much, I didn't document my important uses. If I had access to their stylesheet *prior* to the browser's minification then it'd be easier to know, but some of their stuff was pretty specific so hard to beat in value. There's also a lot of race condition css loading where variables changed wouldn't apply upstream early enough so I have to see what I can do.

uncenter:
We should avoid using !important as much as possible. I guess now you should try removing it everywhere and see what breaks?

PR #2106, opened 2026-01-27
feat(lib/lib): use more accurate COBYLA CSS filters

WalkQuackBack:
Updates the Catppuccin Filters in the standard library to use the COBYLA CSS filters in https://github.com/catppuccin/whiskers/pull/116


ozwaldorf:
image

Comparison between the true colors, COBYLA filters, and the previous filters

PR #2108, opened 2026-01-29
fix(codeberg): header elements

ionawr:
## 🔧 What does this fix? 🔧

Fixes an issue where svg elements were not coloured correctly in `ui.secondary.menu`.

- Before
  image
- After
  image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/codeberg/catppuccin.user.less
+++ b/styles/codeberg/catppuccin.user.less
@@ -49,16 +49,6 @@
     --color-link: @blue;
     --color-footer-text: @text;
 
-    :is(#navbar > .menu, #navbar > .menu .menu) > a,
-    :is(#navbar > .menu, #navbar > .menu .menu) > button,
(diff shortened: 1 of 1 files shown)


uncenter:
Looking at Codeberg's source CSS for that/those elements, it looks like the root cause is:

```css
:is(#navbar > .menu, #navbar > .menu .menu) > a, :is(#navbar > .menu, #navbar > .menu .menu) > button, :is(#navbar > .menu, #navbar > .menu .menu) > details > summary {
	color: #fff;
}
```

We can simplify this a bit but we should probably use selector(s) more similar to this.

ionawr:
@uncenter thanks for the review, updated the code to follow the suggested selector.
i can confirm that this version works well for both light and dark themes.

image
image

uncenter:
Sorry, adjusted the selector further to wrap everything under `#navbar` and also take a slightly different approach to theming the navbar based on Codeberg's original styles.

PR #2111, opened 2026-01-30
fix(github): misc.

ninetailedtori:
## 🔧 What does this fix? 🔧

- Unthemed stripes in diff deletion squares: #fff9 (upstream) -> @surface0 (catppuccin)
image
- Inbox zero icon:
image

## 🗒 Checklist 🗒


uncenter:
> For reproduction page, this page works to see deletion squares within the file header:

Provide a URL in the future.

This solution targets an autogenerated class. Please do not use this, as outlined in our contributing guidelines and as mentioned repeatedly on your pull requests as of late. Also, as the majority of this fraction of a pixel element _is_ themed this seems like something not worth fixing as it is barely noticeable and annoying to theme given the available classes. Regardless, avoid autogenerated prone-to-breakage classes.

uncenter:
These changes are niche and trivial and don't seem worth the additional weight. I'd argue we can't reasonably theme every complicated SVG graphic that GitHub uses. The diff deletion square are also a miniscule page element and minor issues are not noticable.

PR #2113, opened 2026-01-31
fix(codeberg): missing elements

ninetailedtori:
## 🔧 What does this fix? 🔧

- Fixes forgejo being overwritten and as such appearing as a squashed codeberg logo. See footer of https://codeberg.org/librewolf for an example, but this issue happens on ALL repos. Looking at the "powered by" logo, which should be "forgejo", which with my fix now is.
- Fixes codeberg icons.
- Fixes navbar race-condition.
- Adds support for entire [homepage](https://codeberg.org/about).

## 🗒 Checklist 🗒

### DIFF
--- a/styles/codeberg/catppuccin.user.less
+++ b/styles/codeberg/catppuccin.user.less
@@ -16,17 +16,25 @@
 ==/UserStyle== */
 
 @import "https://userstyles.catppuccin.com/lib/lib.less";
+// We need CodeMirror lib too!
 
 @-moz-document domain("codeberg.org") {
+  // GITEA
   @import (css)
     url("https://catppuccin.github.io/gitea/theme-catppuccin-@{lightFlavor}-@{accentColor}.css")
     (prefers-color-scheme: light);
   @import (css)
     url("https://catppuccin.github.io/gitea/theme-catppuccin-@{darkFlavor}-@{accentColor}.css")
     (prefers-color-scheme: dark);
-}
 
-@-moz-document domain("codeberg.org") {
+  // CHROMA
+  @import (css)
+    url("https://chroma.catppuccin.com/@{lightFlavor}-chroma-style.css")
+    (prefers-color-scheme: light);
+  @import (css)
+    url("https://chroma.catppuccin.com/@{darkFlavor}-chroma-style.css")
+    (prefers-color-scheme: dark);
+
   [data-theme="codeberg-dark"],
   [data-theme="codeberg-light"],
   [data-theme="codeberg-auto"] {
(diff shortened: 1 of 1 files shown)


styles/codeberg/catppuccin.user.less
```diff
@@ -76,10 +77,14 @@
     #navbar-logo img {
       #codeberg-logo(@crust);
     }
-    .branding img {
+    :not(.powered-by > a) > .branding img {
```
uncenter:
```suggestion
    .branding img[src^="https://design.codeberg.org/logo-kit/"] {
```

Could filter it to logos from their logo kit instead, looks a little cleaner.

styles/codeberg/catppuccin.user.less
```diff
@@ -47,7 +47,8 @@
     --color-nav-hover-bg: darken(@accent, 5%);
     --color-secondary-nav-bg: @surface0;
     --color-link: @blue;
-    --color-footer-text: @text;
+    --color-footer: @mantle;
```
uncenter: What is this for?

styles/codeberg/catppuccin.user.less
```diff
@@ -76,10 +77,14 @@
...
       #codeberg-logo(@crust);
     }
-    .branding img {
+    :not(.powered-by > a) > .branding img {
       #codeberg-logo(@text);
     }
 
+    a[aria-label="Forgejo"] {
+      filter: @text-filter;
+    }
```
uncenter: Is this for an image? This selector doesn't make it clear.

styles/codeberg/catppuccin.user.less
ninetailedtori: Footer background. Default upstream is dark green. Catppuccin is mantle.

styles/codeberg/catppuccin.user.less
ninetailedtori: Let's do that then. I'll make it filter both by logo tag/content instead.

styles/codeberg/catppuccin.user.less
ninetailedtori: Yes, this is the forgejo image, but it's done as a bg image to the link.

uncenter (changes requested):
How much of this is still necessary?

ninetailedtori:
> How much of this is still necessary?

Anything that's still in it is still necessary, I went through and checked against the upstream after fixing that issue.

uncenter:
It doesn't appear that theming the Codeberg home page is worth it - since none of it is broken by our theming either, I think we can leave it be.

I don't know what a/the navbar race condition is.

Codeberg icons? should now already be fixed in #2108.

I would appreciate a separate pull request raised for the Forgejo squashed logo issue.

PR #2114, opened 2026-02-01
fix(reddit): trending carousel cards text

rohanp2051:
## 🔧 What does this fix? 🔧

Fixes black/unreadable text on the Trending Today carousel cards on the Reddit homepage.

The `text-primary-onBackground` class uses `--color-primary-onBackground`, which the theme sets to a dark color. This breaks text that appears over image cards with gradient overlays.

Fix targets elements inside gradient backgrounds and forces white text:

### DIFF
--- a/styles/reddit/catppuccin.user.less
+++ b/styles/reddit/catppuccin.user.less
@@ -66,7 +66,7 @@
     --color-primary-background: @accent !important;
     --color-primary-background-hover: darken(@accent, 5%) !important;
     --color-primary-background-selected: darken(@accent, 5%) !important;
-    --color-primary-onBackground: @text !important;
+    --color-primary-onBackground: @base !important;
     --color-primary-onBackground-selected: @crust !important;
     --color-secondary: @subtext0 !important;
     --color-secondary-hover: @subtext1 !important;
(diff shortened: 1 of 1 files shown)


uncenter:
Can I see a screenshot of these elements before and after? And can you clarify where these elements are meant to be? I've been poking around the site while both logged in and not and haven't seen this issue.

rohanp2051:
Here is what it was looking like before: 
Screenshot 2026-01-31 at 8 11 44 PM
Here is after: 
Screenshot 2026-01-31 at 8 09 28 PM

uncenter:
I still can't locate those cards on my Reddit page, is this something I need to configure/enable/toggle in settings?

rohanp2051:
I think the issue is that your signed in. This is what reddit.com looks like for users that aren't signed in or using any extensions for modification.

uncenter:
Ah thanks I opened in a new clean window and reproduced.

uncenter:
Can we try setting `--color-primary-onBackground` to `@text` directly? Nothing looks immediately broken with that change.

rohanp2051:
Switched to setting --color-primary-onBackground to @text directly and removed the targeted selectors. Also rebased onto latest main.
image

uncenter:
Shoot I'm sorry actually I think I mislead you. As soon as you changed that I noticed this case where this text ("NEW") would previously have been base but now is text and lacks contrast.

CleanShot 2026-02-26 at 09 16 46@2x

Sorry for the hassle, can we go back to your original solution?

PR #2116, opened 2026-02-01
refactor(roblox-devforum): use discourse base

ninetailedtori:
## 🔧 What does this fix? 🔧

Subject to changes, there's still a bit that has to be done:
- do we theme categories?
  - category badges are all specific to the site, so it is likely impossible to do this, especially lightly.
- if we make a library, this *might* be easier
  - especially if we make an official port, this would allow site admins to port it over cleanly on their own upstream sites.
  - even so, making a lib is quite useful, for our own userstyles. something like gitea or our upstream syntax libs or mediawiki lib, for example, where we can simply include it and have the discourse variables set.

Sites used by me to test Discourse compatibility:
- https://discourse.nixos.org/
- https://meta.discourse.org/
- https://devforum.roblox.com/
- https://community.getmailspring.com/

## 🗒 Checklist 🗒

### DIFF
--- a/styles/discourse/catppuccin.user.less
+++ b/styles/discourse/catppuccin.user.less
@@ -30,22 +30,23 @@
     #catppuccin(@darkFlavor);
   }
 
-  #w-rgb(@id,
-  @color) {
+  #w-rgb(@id, @color) {
     --@{id}: @color;
     --@{id}-rgb: #lib.rgbify(@color)[];
   }
 
-  #w-hover(@id,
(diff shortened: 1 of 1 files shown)


ninetailedtori:
Do we wish to support specific site logos and overrides as well? We certainly have space but we should also keep in mind to keep it light still. Maybe only the big popular discourse sites.
We'll add them to the supports section after. I'll add the overrides for discourse and nixos.

PR #2122, opened 2026-02-04
feat(lib): move to std directory, versioned path, documentation

WalkQuackBack:
Implement versioning into the Standard Library.

Checklist before merging:

- [x] Update documentation to reflect new structure 
- [x] Document introduction of new library updates by Userstyles Staff


lib/lib.less
```diff
@@ -1,129 +1,2 @@
...
-    --ctp-overlay0: @overlay0;
-    --ctp-surface2: @surface2;
-    --ctp-surface1: @surface1;
-    --ctp-surface0: @surface0;
-    --ctp-base: @base;
-    --ctp-mantle: @mantle;
-    --ctp-crust: @crust;
-  }
-}
+// Standard library has migrated to a versioned syntax, see <DOCS URL>
```
uncenter: Seems like we should add a docs page in this pull request too?

lib/lib.less
```diff
@@ -1,129 +1,2 @@
...
+@import "https://userstyles.catppuccin.com/lib/std/v1.less";
```
uncenter, line 2: Wonder if it is better to double import like so or add a redirect for our hosting provider between the URLs... the latter does require more coordination with staff cc @sgoudham

WalkQuackBack: As far as I know, we use GitHub Pages as our "hosting provider", and it doesn't support redirects. But if it can be done in a clean way, it would definitely be better than the double import.

uncenter: Fair enough, I forgot GH Pages doesn't support that. Well, we can do the double import for now.

lib/lib.less
uncenter: Do we need to provide a URL for the migration change itself or just the library (modules) documentation in general?

docs/src/content/docs/contributing/reference/libraries/modules/index.md
uncenter, line 1: Is it worth structuring this into some kind of folder?

lib/lib.less
WalkQuackBack: I think the latter is okay considering the small amount of libraries, currently.

WalkQuackBack: Actually, there are hundreds of open PRs built on the current version of the libraries. We should provide a migration path in documentation.

uncenter: As in pull requests that import userstyles.catppuccin.com/lib/lib.less? For the original standard library introduction I just manually went through and changed the URLs. We could do that, or just leave suggestions, etc. I personally don't think adding a documentation page for it is necessary but happy to discuss further if you feel strongly.

uncenter: I added a changelog section and an entry for v1 that we can link to here!

uncenter:
Ah hmm. I think the imports are failing because we aren't replacing them in the lib content itself. Will try to think of a fix.

uncenter:
> Ah hmm. I think the imports are failing because we aren't replacing them in the lib content itself. Will try to think of a fix.

To clarify for myself - the imports of userstyles.catppuccin.com in the lib *shim* aren't being replaced because we are only patching URLs in the main userstyle content itself. The better approach would be to intercept fetch requests to that host and return responses to the local file system, but I'm having trouble implementing that in Deno.

uncenter:
This is looking good to me with a new documentation structure and updated for v1, v1 moved from lib/lib.less to lib/std/v1.less, and the linting has been fixed to match the local repository instead of the remote website in #2292.

docs/src/content/docs/contributing/reference/libraries/modules/index.md
```diff
@@ -0,0 +1,28 @@
...
+  label: Overview
+---
+
+Library modules consist of collections of mixins which target design systems or commonly used libraries.
+
+They follow a standard structure and should be built to be modular and reusable across several userstyles.
+
+## Structure
+
+The [standard library](https://github.com/catppuccin/userstyles/blob/main/lib/std/v1.less) is a good reference as to how libraries can be structured.
```
WalkQuackBack:
```suggestion
The [standard library](/contributing/reference/libraries/standard/) is a good reference as to how libraries can be structured.
```
There may be multiple versions of standard library, so link to the internal documentation page which contains a link to the source code.

docs/src/content/docs/contributing/reference/libraries/modules/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+All libraries are in lib/ with subdirectories per library name, and versioned files (e.g., `v1.less`, `v2.less`, etc).
+
+They must have a top level mixin in the format `#__libraryname`. All constants and variables should be under a nested `.base` mixin, used like `#__libraryname.base()` under `#lib.palette()`.
+
+You may include other mixins and constants, but they must be under the top level mixin for scoping purposes.
```
WalkQuackBack:
```suggestion
Other mixins and constants may be included, but they must be under the top level mixin for scoping purposes.
```

uncenter (dismissed):

docs/src/content/docs/contributing/reference/libraries/modules/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+They follow a standard structure and should be built to be modular and reusable across several userstyles.
+
+## Structure
+
+The [standard library](https://github.com/catppuccin/userstyles/blob/main/lib/std/v1.less) is a good reference as to how libraries can be structured.
+
+All libraries are in lib/ with subdirectories per library name, and versioned files (e.g., `v1.less`, `v2.less`, etc).
+
+They must have a top level mixin in the format `#__libraryname`. All constants and variables should be under a nested `.base` mixin, used like `#__libraryname.base()` under `#lib.palette()`.
```
WalkQuackBack:
```suggestion
They must have a top level mixin in the format `#__libraryname`. All constants and variables should be under a nested `.base` mixin. The nested mixin is used like `#__libraryname.base()` after all other standard library statements.
```
Removing the direct "under `#lib.palette()`" as it doesn't apply correctly when the `#lib.css-variables()` is used.

docs/src/content/docs/contributing/reference/libraries/modules/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+You may include other mixins and constants, but they must be under the top level mixin for scoping purposes.
+
+## Creation
+
+Identify what your library module will achieve. Are you targeting a design system, framework, or a library?
+
+If a design system or framework, is it specific to one userstyle or reusable across multiple? What are the popularity metrics of it? Is it worth the maintenance burden for the benefit it provides in DRY?
```
WalkQuackBack:
```suggestion
If a design system or framework, is it specific to one userstyle or reusable across multiple? What are the popularity metrics of it? Is the increased maintenance workload worth the benefit it provides in DRY (Don't Repeat Yourself)?
```
"burden" has a negative connotation, so avoiding using that.
Readers may not be familiar with what DRY means, so added an acronym expansion.

docs/src/content/docs/contributing/reference/libraries/modules/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+If it is a web library, you should check if there is an [existing port](http://catppuccin.com/ports) for it. You may be able to use the methods outlined in [Syntax Highlighting](/contributing/guides/syntax-highlighting/) to import the theme for the library in question. However, if there is no CSS support it may be necessary to upstream the changes needed for it to work.
```
WalkQuackBack:
```suggestion
If the target is a web library, check if there is an [existing port](http://catppuccin.com/ports) for it. The methods outlined in [Syntax Highlighting](/contributing/guides/syntax-highlighting/) may be used to import the theme for the library in question. Some ports may need changes upstreamed to work for userstyles purposes; in these cases, please coordinate with userstyles staff to determine the most suitable approach.
```

Reiterating "the target" and making some conciseness edits. Don't really like how I reworded the last sentence, some more work could be done to it.

docs/src/content/docs/contributing/reference/libraries/standard/index.mdx
```diff
@@ -0,0 +1,34 @@
...
+import { Badge } from '@astrojs/starlight/components';
+
+
+| Versions                                                                          | Source Code                                      | Hosted Library |
+| --------------------------------------------------------------------------------- | ------------------------------------------------ | -------------- |
+| [v1 <Badge text="Latest" variant="success" style={{ marginLeft: '0.5rem' }} />](/contributing/reference/libraries/standard/v/v1/)       | [`lib/std/v1.less`](https://github.com/catppuccin/userstyles/blob/main/lib/std/v1.less) | https://userstyles.catppuccin.com/lib/std/v1.less |
+
+---
+
+The standard library is a collection of mixins to simplify and prevent redundant code in userstyles. In userstyles, the standard library is imported at the top of the file (alongside with other [library modules](/contributing/reference/libraries/modules/)):
```
WalkQuackBack:
```suggestion
The standard library is a collection of mixins to deduplicate and simplify code in userstyles. It is imported in every userstyle, alongside with other [library modules](/contributing/reference/libraries/modules/) if applicable.
```

docs/src/content/docs/contributing/reference/libraries/standard/index.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+```less
+@import "https://userstyles.catppuccin.com/lib/std/v1.less";
+```
```
WalkQuackBack: Should we include this here? It is also part of the template.

docs/src/content/docs/getting-started/faq.md
```diff
@@ -12,3 +12,19 @@ Sorry to hear that! Please [open an issue](https://github.com/catppuccin/usersty
...
+  - _Chrome/Chromium_: the version should be the first entry in the table that appears. There is a copy to clipboard button after the version number which you can use.
+  - _Firefox_: the version is under the "Application Basics" section in the row labeled "Version".
+
+| Browser        | Version             |
+| -------------- | ------------------- |
+| Chromium       | `chrome://version/` |
+| Firefox        | `about:support`     |
+| Opera (GX)     | `opera://about/`    |
+| Microsoft Edge | `edge://version/`   |
+| Arc            | `arc://version/`    |
```
WalkQuackBack, line 30: Was this intended to be in this PR?

docs/src/content.config.ts
WalkQuackBack, line 1: Looks good :)

docs/astro.config.mjs
```diff
@@ -57,37 +58,22 @@ export default defineConfig({
...
         starlightImageZoom(),
+        starlightAutoSidebar(),
       ],
     }),
   ],
+  redirects: {
+    "/contributing/standard-library/":
+      "/contributing/reference/libraries/standard/",
+    "/reference/browsers/":
+      "/getting-started/faq/#what-version-of-my-browser-am-i-using",
```
WalkQuackBack, line 77: Same comment as https://github.com/catppuccin/userstyles/pull/2122/changes#r3619182454

scripts/generate/templates/userstyle-issue.yml
```diff
@@ -36,7 +36,7 @@ body:
     type: input
     attributes:
       label: Browser version(s) this issue occurs on
-      description: "To get the browser version, see https://userstyles.catppuccin.com/reference/browsers."
+      description: "To get the browser version, see https://userstyles.catppuccin.com/getting-started/faq/#what-version-of-my-browser-am-i-using."
```
WalkQuackBack, line 39: Same comment as https://github.com/catppuccin/userstyles/pull/2122/changes#r3619182454

WalkQuackBack (commented):
Made some minor edits and moved some links around, but overall looks good! Thanks for the work documenting all of this ;)

docs/src/content/docs/contributing/reference/libraries/modules/index.md
uncenter: Did some more rewording to what I think your intended meaning is?

docs/src/content/docs/contributing/reference/libraries/standard/index.mdx
uncenter: I think it is relevant to show that this is what importing it looks like, otherwise it may be a little nebulous until you see it in practice.

docs/src/content/docs/getting-started/faq.md
uncenter: I reorganized the reference section by moving some stuff into the FAQ to make room for this new reference libraries section. I can move it out but I don't think its worth blocking & the time.

docs/src/content/docs/contributing/reference/libraries/standard/index.mdx
WalkQuackBack: What about this? Though it would have to be updated every library bump.

```suggestion

## Importing

```less
@import "https://userstyles.catppuccin.com/lib/std/v1.less";
```
```

docs/src/content/docs/getting-started/faq.md
WalkQuackBack: Sounds good to me leaving it in, then. Thanks for clarifying! :+1:

docs/src/content/docs/contributing/reference/libraries/modules/index.md
WalkQuackBack: Yes, that works!

docs/src/content/docs/contributing/reference/libraries/standard/index.mdx
uncenter: I'm fine with updating it on every bump just so it is up to date and helpful. This looks good to me.

uncenter:
Good to go?

WalkQuackBack:
Did one more review, looks good! @uncenter you may approve

uncenter (approved):
LGTM!

ParasailNumerous (approved):

WalkQuackBack:
CI passed, LGTM

PR #2124, opened 2026-02-04
refactor(lib/m3): integrate versioning

WalkQuackBack:
Integrates versioning into the M3 library with a backward compatibility shim


lib/m3.less
uncenter, line 1: No point in adding this since it hasn't been published yet right?

WalkQuackBack: The WIP PRs rely on this path; they need to be updated to use the versioned format before this can be properly removed.

uncenter: I think we can just remove the shim and update the dependent pull rqeuests.

PR #2136, opened 2026-02-16
fix(github): feed link color

a3chron:
## 🔧 What does this fix? 🔧

Small fix: color of links in feed, they sometimes are not styled, got a hardcoded color apparently

(It's not all links, I think it's just the ones in a md code block, not sure)

Before:
image

After:
image

## 🗒 Checklist 🗒

### DIFF
--- a/styles/github/catppuccin.user.less
+++ b/styles/github/catppuccin.user.less
@@ -463,7 +463,7 @@ regexp(
     --outline-focus: @blue solid 2px;
 
     /* Feed Links */
-    .feed-item-content a tt {
+    .feed-item-content a code, .feed-item-content a tt {
       color: @accent;
     }
 


uncenter:
Can you share a reproduction URL?

a3chron:
Could be a little hard, as the issue seems to be only in the feed (bc of that `.feed-item-content`), and on the actual release page it works again, but you could check in your own github feed, there may be changelogs of repos you starred with similar links to commits

a3chron:
You should be able to filter your feed to only show releases, so you might be able to find sth quicker like that, sry for the inconvenience 🙃

uncenter:
Oh yep thanks I see that. Annoying override from `billing.scss` apparently (are they shipping source maps...?).

styles/github/catppuccin.user.less
```diff
@@ -462,6 +462,11 @@ regexp(
     --shadow-floating-legacy: 0px 6px 12px -3px @crust, 0px 6px 18px 0px @crust;
     --outline-focus: @blue solid 2px;
 
+    /* Feed Links */
+    .feed-item-content a tt {
```
uncenter:
```suggestion
    .feed-item-content a code, .feed-item-content a tt {
```

Looks like they use this selector that covers more so we can use the same thing to catch all cases.

a3chron:
Thanks, didnt see that, commited the suggestion :)

PR #2127, opened 2026-02-05
chore(ISSUE_TEMPLATE): add form fields for stylus and userstyle versions

uncenter:
Rewords grammar to be more concise and adds addition fields to assist with reproducing userstyle bugs.

PR #2189, opened 2026-03-24
chore(ISSUE_TEMPLATE): various improvements

uncenter:
Did it properly this time - https://github.com/catppuccin/userstyles/pull/2127 got reset because I forgot that issue templates are generated... oops!

Same changes, plus some meta changes with renaming the links and issue template titles.

Also added ids to the form fields. This will allow us to prefill issue templates - I am thinking that we can set the supportURL in the UserCSS to include the `userstyle_version` field, and possibly the `userstyle` field as well. Would be great if Stylus had some way to prefill browser (`browser_version`) and Stylus versions (`stylus_version`) as well in the future. This concept also has relatively minimal outcomes given that the only use of supportURL is for this small "Feedback" link in this part of the popup:

CleanShot 2026-03-24 at 16 12 30@2x

I might open an issue for making it more useful.


uncenter:
Tracking improvements in Stylus at https://github.com/openstyles/stylus/issues/2061.

PR #2220, opened 2026-04-17
feat(fabric-docs): init

scarcekoi:
**Website**: Fabric Docs

**URL**: https://docs.fabricmc.net

**Description**: Docs for the Minecraft mod loader, Fabric.

preview

## 🗒 Checklist 🗒

---

- [x] I have read and followed Catppuccin's [userstyle submission guidelines](https://userstyles.catppuccin.com/contributing/creating-userstyles/).
- [x] I have made a new directory underneath `/styles/<name-of-website>`.
  - [x] I have named the userstyle `catppuccin.user.less` within the new directory.
  - [x] I have followed the [template](https://github.com/catppuccin/userstyles/blob/main/template/catppuccin.user.less) and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have updated the [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml) file with information about the new userstyle.

### DIFF
--- a/scripts/userstyles.yml
+++ b/scripts/userstyles.yml
@@ -75,6 +75,7 @@ collaborators:
   - &koibtw koibtw
   - &TheAnonymousCrusher TheAnonymousCrusher
   - &42willow 42willow
+  - &scarcekoi scarcekoi
 
 userstyles:
   advent-of-code:
@@ -296,6 +297,12 @@ userstyles:
     categories: [social_networking]
     color: peach
     current-maintainers: [*ryanccn]
+  fabric-docs:
+    name: Fabric Docs
+    link: https://docs.fabricmc.net
+    categories: [wiki]
+    color: blue
+    current-maintainers: [*scarcekoi]
   freedesktop:
     name: freedesktop.org
     link: https://www.freedesktop.org
--- a/styles/fabric-docs/catppuccin.user.less
+++ b/styles/fabric-docs/catppuccin.user.less
@@ -29,57 +29,92 @@
     #lib.palette();
     #lib.defaults();
 
-    --vp-c-white:@crust;
-    --vp-c-black:@crust;
+    /* Fix button text color */
+    --vp-button-brand-text: @crust !important;
+    --vp-button-brand-hover-text: @crust !important;
+    --vp-button-brand-active-text: @crust !important;
 
-    --vp-c-gray-1:@surface0;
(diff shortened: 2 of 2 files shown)


uncenter:
This site uses Vitepress, as seen in the `--vp` for Vitepress variables. We have a theme for it at https://github.com/catppuccin/vitepress which we should try to incorporate.

scarcekoi:
I pulled the colours from the official port and used those.

uncenter:
This should probably be a userstyle for Vitepress itself which applies to Fabric's documentation, like how our MdBook port works.

scarcekoi:
Well fabric needed the override for the tip soft colour because on the docs it's the accent.
I think it would be good to have a vitepress userstyle, would it be possible to add an override on specific websites?

uncenter:
The other option is to have a library version of Vitepress that we can import from here.

scarcekoi:
> The other option is to have a library version of Vitepress that we can import from here.

Would this be done on the official port? How could I do that? I'd be happy to work on it.

42willow:
the vitepress port uses whiskers at the moment so it'd need to be changed there-

i'm not sure how to do that though because at the moment we're just [extending the default theme](https://vitepress.dev/guide/extending-default-theme) by importing the css file in `.vitepress/theme/index.js` and that somehow overrides the default css. i'm not sure how this would be done dynamically (or whatever the terminology is)

although it's probably unlikely that vitepress will release any breaking changes to it's css variables any time soon (if at all) - so the current approach (and renaming the userstyle to vitepress) might be better :)

uncenter:
> the vitepress port uses whiskers at the moment so it'd need to be changed there-

We can use e.g. https://github.com/catppuccin/vitepress/blob/main/theme/frappe/blue.css can we not?

42willow:
> > the vitepress port uses whiskers at the moment so it'd need to be changed there-
> 
> 
> 
> We can use e.g. https://github.com/catppuccin/vitepress/blob/main/theme/frappe/blue.css can we not?

Like `@import`? For some reason I thought this could be blocked by some sites but I'm not sure-

Oops sorry person who's username is import I'm on mobile and forgot to code block 🙃

uncenter:
> Like `@import`? For some reason I thought this could be blocked by some sites but I'm not sure-

It could be but we can just test it.

PR #2235, opened 2026-04-26
feat(codeberg): accessible header, landing page

ninetailedtori:
## 🗒 Checklist 🗒

  - NEVERRRRRRRRRRRRRRRRRRRRRRRRRRR [angry anti-ai noises]

## 🔧 What does this fix? 🔧

- [x] Makes the navbar mantle, with hover buttons as accent, it's cleaner and more accessible to read imo (less violently bright on accents like rosewater).
- [x] Adds the codeberg landing page theming. [Codeberg | About](https://codeberg.org/about)
- [x] Adds isWarning class for isSigned shasums which aren't verified.

## Troubles with this PR?

> Yes, actually! Here's a list of pages I _haven't_ themed, and WHY:

- so, docs.codeberg.org, relies a ton on *halfmoon, which is a theme loaded through bootstrap*. that's honestly gonna be quite painful to theme, without a dedicated library, which is gonna require a lot of work, because there's about 500 variables here that are all bound and tangled, and a lot are just duplicates that navigate to other vars, etc. **See halfmoon.css on that site.** This is writable, but we'll have to write a lib, which will then define the 500 or so variables.
  - As a secondary, uses PrismJS for syntax-highlighting.
- blogs.codeberg.org uses...*scoped halfmoon, scoped to its own codeberg-design module*, which...umm, somehow defines a whole new 1500 variables. That's just not gonna be fun. **See codeberg.css on that site.** This just isn't feasible, so either we just sloppily define what we can see, and patch as we go along, or we just ignore the website entirely. It's not a major subdomain that people want afaik?
  - As a secondary, uses Pygments for syntax-highlighting.
- donate.codeberg.org is the lightest, and uses pico, which you can see, **lies in pico.min.css which you can see on the site**. It's not hard to theme, but the *correct* thing to do in accordance with our own userstyle/lib modularisations, is to write a lib for pico, then import it. This one does have a very light solution but not something we should care about until we get a pico library, because it's not worth doing yet, and not something major, just like blogs.

### DIFF
--- a/styles/codeberg/catppuccin.user.less
+++ b/styles/codeberg/catppuccin.user.less
@@ -14,17 +14,18 @@
 @var select darkFlavor "Dark Flavor" ["latte:Latte", "frappe:Frappé", "macchiato:Macchiato", "mocha:Mocha*"]
 @var select accentColor "Accent" ["rosewater:Rosewater", "flamingo:Flamingo", "pink:Pink", "mauve:Mauve*", "red:Red", "maroon:Maroon", "peach:Peach", "yellow:Yellow", "green:Green", "teal:Teal", "blue:Blue", "sapphire:Sapphire", "sky:Sky", "lavender:Lavender", "subtext0:Gray"]
 ==/UserStyle== */
+
 @import "https://userstyles.catppuccin.com/lib/lib.less";
 
 @-moz-document domain("codeberg.org") {
-  // GITEA
+  /* GITEA */
   @import (css)
     url("https://catppuccin.github.io/gitea/theme-catppuccin-@{lightFlavor}-@{accentColor}.css")
     (prefers-color-scheme: light);
   @import (css)
     url("https://catppuccin.github.io/gitea/theme-catppuccin-@{darkFlavor}-@{accentColor}.css")
     (prefers-color-scheme: dark);
-  // We need CodeMirror lib too!
+  /* @import url(CM_LIB_URL); */
 }
 
 @-moz-document domain("blog.codeberg.org") {
@@ -35,9 +36,11 @@
   @import url("https://prismjs.catppuccin.com/variables.important.css");
 }
 
-// @-moz-document domain("donate.codeberg.org") {
-//   @import url(PICO_LIB_URL);
-// }
+/*
+ * @-moz-document domain("donate.codeberg.org") {
+ *   @import url(PICO_LIB_URL);
+ * }
+ */
 
 @-moz-document domain("codeberg.org") {
   [data-theme="codeberg-dark"],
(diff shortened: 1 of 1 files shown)


uncenter:
As I've stated before, I'm not quite sure theming the donation and blog pages are the best use of our time and maintenance especially given their length and complexity. Can we remove those changes and just focus on the "accessible header"?

ninetailedtori:
> As I've stated before, I'm not quite sure theming the donation and blog pages are the best use of our time and maintenance especially given their length and complexity. Can we remove those changes and just focus on the "accessible header"?

I did remove them. Just the barebones rn, please check. Accessible header, and landing page (which was very cheap to maintain.)

ninetailedtori:
There are a couple single-line imports, because the docs and blogs each use a library, one pygments, one prism, and it's quite lightweight for us not to care about that and to just have a one-line import for each domain.

styles/codeberg/catppuccin.user.less
```diff
@@ -18,37 +18,71 @@
...
+    }
+
+    .ui.secondary.menu .dropdown.item:focus,
+    .ui.secondary.menu .dropdown.item:active,
+    .ui.secondary.menu a.item:focus,
+    .ui.secondary.menu a.item:active,
+    .ui.secondary.menu a.active.item:focus,
+    .ui.secondary.menu a.active.item:active {
+      background-color: var(--color-nav-active-bg);
+    }
```
isabelroses: this and codeblocks like it could surely be minified

ninetailedtori: Yeah lemme bloop that minif colon three style :3

styles/codeberg/catppuccin.user.less
```diff
@@ -18,37 +18,74 @@
...
     }
   }
 
   #catppuccin(@flavor) {
     #lib.palette();
-    // #lib.defaults(); // Not needed, themed in upstream Gitea port.
+    /*
+     * #lib.defaults();
+     * This isn't needed, themed in upstream Gitea port.
+     */
```
uncenter: These changes are all unnecessary or secondary.

uncenter (changes requested):

styles/codeberg/catppuccin.user.less
ninetailedtori: I was attempting to up the documentation on the port to make it easier to maintain, though. is that not something that's wanted?

ninetailedtori: Well, removed anyway.

isabelroses (approved):
lgtm

PR #2238, opened 2026-04-29
chore(renovate): enable pin github actions digests

uncenter:
This is a recommended security practice for GitHub Actions these days. I think after merging Renvoate will start opening pull requests pinning each action to its digest.

See https://docs.renovatebot.com/modules/manager/github-actions/#digest-pinning-and-updating.


sgoudham (approved):
O

Issue #2222, opened 2026-04-19
YouTube: Maintainability issues


WalkQuackBack:
### Is there an existing issue outlining your problem?

- [x] I have searched the existing issues and they do not solve my problem.

### What userstyle are you seeing the problem on?

lbl:youtube

### Describe your problem.

As mentioned at https://github.com/catppuccin/userstyles/issues/2219#issuecomment-4269138949, YouTube as it migrates towards a styling solution matching modern Google Search is become almost impossible to maintain.

This is a subsection of some of the new hashed CSS variables.

```css
[dark] {
    --t7f4f2c6d54836ce0: rgba(255,255,255,0.1);
    --t7d8b8e5ee291aec0: rgba(0,0,0,0.05);
    --t4726c68ef6d7b08f: rgba(0,63,13,0.6);
    --tf0a0d1154827847f: rgba(124,41,0,0.6);
...
```

```css
.ytChipShapeActive {
    background-color: var(--t1405e70a39276293);
    color: var(--t6216186c28b3834b);
}
```

```css
html:not(.style-scope) {
    --paper-dialog-background-color: var(--t518e925f61bdcb91);
    --paper-listbox-background-color: var(--t518e925f61bdcb91);
    --paper-listbox-color: var(--tffc2fd3a644f6275);
}
```

While these updates also introduced new YouTube system variables (`--yt-spec-*`), the hardcoded variables are not being bound to these and instead are being resolved at SCSS build time.

It should be determined how we should move forward from this.

Related: #2219, #2217, #2212

### Browser version(s) this issue occurs on

N/A

### Stylus extension version

N/A

### Installed version of the userstyle

N/A

nyxiereal:
we are so cooked, ctp broke for me on youtube a few days ago. now i know why... classic

uncenter:
@nyxiereal please see #2219.

spock-22:
i tried to patch the problem for the moment. adding these lines to the bottom of the userstyle themes at least the majority of the youtube page for me:
```
[dark] {
  --t3e41d7b17b187f69: #1e1e2e !important;
  --ta889dfda9605a358: #181825 !important;
  --t7f4f2c6d54836ce0: #313244 !important;
  --t08a7c6c176cbc5c2: #313244 !important;
  --tffc2fd3a644f6275: #cdd6f4 !important;
  --paper-toggle-button-checked-button-color: #b4befe !important;
  --tf3fc855af2285f5f: #313244 !important;
[...]

Issue #2219, opened 2026-04-16
Youtube Unthemed


fe50c:
### Is there an existing issue outlining your problem?

- [x] I have searched the existing issues and they do not solve my problem.

### What userstyle are you seeing the problem on?

lbl:youtube

### Describe your problem.

Youtube is entirely unthemed by userstyle - likely just a continuation of the updates they made a few days ago.
(every page I thought to test is affected)

### Browser version(s) this issue occurs on

LibreWolf 148.0.2-2

### Stylus extension version

Stylus 2.3.19

### Installed version of the userstyle

v2026.04.14

fe50c:
Interestingly the loading modal seems themed, but when the homepage fully loads it's entirely unthemed (attached below)

<img width="2560" height="1440" alt="Unthemed homepage" src="https://github.com/user-attachments/assets/03d9bf3c-e467-4def-8271-05ddb3db4080" />

hajosattila:
same problem here...

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/66e0f062-93d0-47ec-99d9-543b5f494e29" />

uncenter:
YouTube's most recent update is introducing more and more unmaintainable variables. We won't be able to release an immediate fix. My best suggestion is to just turn off the userstyle for now and subscribe to notifications on this issue on the panel on the right for when we have updates.

PR #2245, opened 2026-05-13
fix(youtube): new variables

d-haas:
[Fix] Added new variables that were included to the website since the repository's last update

### DIFF
--- a/styles/youtube/catppuccin.user.less
+++ b/styles/youtube/catppuccin.user.less
@@ -71,6 +71,19 @@
         --yt-sys-color-baseline--overlay-background-solid: @crust;
         --yt-sys-color-baseline--overlay-text-primary: @text;
 
+        /* New newest variables */
+        --yt-sys-color-baseline--overlay-solid-background-inverse: @base;
+        --yt-sys-color-baseline--solid-background-inverse: @base;
+        --yt-sys-color-baseline--static-black: @base;
+        --yt-sys-color-baseline--scrim-background-gradient-1: @overlay0;
+        --yt-sys-color-baseline--scrim-background-gradient-2: @surface2;
+        --yt-sys-color-baseline--scrim-background-gradient-3: @surface1;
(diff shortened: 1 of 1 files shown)


styles/youtube/catppuccin.user.less
```diff
@@ -71,6 +71,19 @@
...
+        --yt-sys-color-baseline--solid-background-inverse: @base;
+        --yt-sys-color-baseline--static-black: @base;
+        --yt-sys-color-baseline--scrim-background-gradient-1: @overlay0;
+        --yt-sys-color-baseline--scrim-background-gradient-2: @surface2;
+        --yt-sys-color-baseline--scrim-background-gradient-3: @surface1;
+        --yt-sys-color-baseline--scrim-background-gradient-4: @surface0;
+        --yt-sys-color-baseline--scrim-background-gradient-5: @base;
+        --t45b2a38314924357: @base;
+        --t6216186c28b3834b: @base;
+        --t3e41d7b17b187f69: @base;
```
isabelroses, line 85: these values are not the same as mine, so my background remains unthemed. seems youtube is doing autogenerated css a/b testing once again

d-haas: I tried to refresh the page and delete the cookies to make sure that wasn't the case, sadly it apparently is :(.

Have your values got similar names or they are totally different?

WalkQuackBack: > these values are not the same as mine, so my background remains unthemed. seems youtube is doing autogenerated css a/b testing once again

It's appears to be the same SSR pipeline as Google Search, so the variables change based on region, A/B tests, and plenty other variables.

uncenter: Since these are autogenerated variables, we cannot reliably theme them. See #2222 and #2219. Let's remove these variables from the pull request and focus on the non-autogenerated ones instead :)

alba4k:
immagine
the top bar remains unthemed in the home screen, but the rest looks good

PR #2249, opened 2026-05-19
fix(chatgpt): fix background

billy4479:
(This is my first PR to this repo, hope everything is fine, let me know!)

## 🗒 Checklist 🗒

- [x] I have read and followed Catppuccin's [userstyle contributing guidelines](https://userstyles.catppuccin.com/contributing/).

- [ ] I used AI (or AI-assistance) for this change.

## 🔧 What does this fix? 🔧

This fixes the background not being themed some times by overriding the `black` variable as well.

## 🔍 Reproduction Steps 🔍

For me, going to chatgpt.com resulted in the background being `#000` instead of the correct color.

| Before | After |
| ------ | ----- |
| image |
image |

PR #2250, opened 2026-05-19
fix(chatgpt): increase selector specificity to resist variable overrides

dertieran:
## 🗒 Checklist 🗒

## 🔧 What does this fix? 🔧

Increases selector specificity in the userstyle so variables are not overridden.
This addresses the same issue as #2249, but fixes it by raising selector specificity rather than changing variables.
Both approaches are acceptable, but this should hopefully also withstand future ChatGPT changes.

## 🔍 Reproduction Steps 🔍

1. Visit https://chatgpt.com/
2. Confirm Catppuccin colors are consistently applied.

Just for reference, this is the selector/variable that broke the background color *(again)*.

```css
html.dark[data-oled], html.dark[data-oled]:not(:where(.light,.light *)) {
    --main-surface-primary: var(--black)
}
```

### DIFF
--- a/styles/chatgpt/catppuccin.user.less
+++ b/styles/chatgpt/catppuccin.user.less
@@ -80,6 +80,10 @@
       // chat input prompt border
       --shadow-color-1: fade(@mantle, 50%);
       --shadow-color-2: fade(@text, 80%);
+      
+      .__menu-item[data-sidebar-item] {
+        --menu-item-highlighted: @surface0;
+      }
 
       .__menu-item[data-color="danger"] {
         --menu-item-highlighted: fade(@red, 20%);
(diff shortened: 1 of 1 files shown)


uncenter:
> This addresses the same issue as #2249, but fixes it by raising selector specificity rather than changing variables.
> Both approaches are acceptable, but this should hopefully also withstand future ChatGPT changes.

I was originally going to go with #2249 as the more idiomatic approach, given that by theming the `--black` variable we fix the background problem without introducing the universal selector, but I did some reading on performance and specificity and I don't think this is a bad option. Let me push some other fixes to this branch and we can go ahead here.

uncenter (approved):

uncenter:
Added a patch to fix the sidebar menu items being broken from the latest update. This LGTM!

PR #2265, opened 2026-06-29
fix(whatsapp-web): #side background gaps

teohhanhui:
## 🗒 Checklist 🗒

## 🔧 What does this fix? 🔧

Fixes some gaps in the background inside the `#side` container

## 🔍 Reproduction Steps 🔍

| Before | After |
| ------ | ----- |
| WhatsApp Web (before) | WhatsApp Web (after) |

### DIFF
--- a/styles/whatsapp-web/catppuccin.user.less
+++ b/styles/whatsapp-web/catppuccin.user.less
@@ -115,6 +115,8 @@
     #WDS(modal-backdrop-solid, @crust);
 
     #WDS(surface-default, @mantle);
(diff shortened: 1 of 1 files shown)


teohhanhui:
The original stylesheet from the site already uses `var(--WDS-surface-default)` for the 2 variables, but it doesn't end up being resolved to *our* `--WDS-surface-default` variable.

WalkQuackBack (approved):
I feel like this is an okay solution in the meantime, but long-term we should try to look into the cascade order more to see why it's not picking up our WDS-surface-default

PR #2267, opened 2026-07-02
fix(duckduckgo): prevent header logo from tiling on mobile

### LINKED ISSUE #2257: DuckDuckGo logo repeated on mobile

### What userstyle are you seeing the problem on?

lbl:duckduckgo

### Description

On mobile, the DDG logo at the top of the page is repeated several times. Reproduced on desktop in Firefox using the Galaxy Note 9 device emulation in dev tools.

### Relevant screenshots

Image

### Browser version(s) this issue occurs on

Firefox 151.0.2

### Stylus extension version

Stylus 2.3.30

### Installed version of the userstyle

2026.03.18

### Any additional comments?

_No response_

SakethKanchi:
### Problem

On mobile, the DuckDuckGo header logo is repeated several times (reported in #2257). The `.header__logo` / `.js-logo-ddg` rule sets a `background-image` (the Catppuccin SVG logo) but never sets `background-repeat`, so on the taller/wider logo container used on mobile the SVG tiles instead of showing once.

### Fix

Add `background-repeat: no-repeat !important;` to the header logo selectors. A logo background should never tile, so this is safe on desktop as well (where the container happened to match the image size, hiding the bug).

### Testing

- `deno task lint duckduckgo` passes.
- Single-line, scoped change to the existing logo rule; no palette/variable changes.

Closes #2257


uncenter:
Did you use LLM-assistance for this change? That needs to be disclosed in contributions, as per our pull request template which you do not appear to have followed.

SakethKanchi:
> Did you use LLM-assistance for this change? That needs to be disclosed in contributions, as per our pull request template which you do not appear to have followed.

My bad on the template, I wrote my own description and didn't realize the checkbox was there. The fix itself is just the one line I added myself, but I did use AI to help write up the description.

styles/duckduckgo/catppuccin.user.less
```diff
@@ -1081,6 +1081,7 @@
...
       );
       background-image: url("data:image/svg+xml,@{svg}") !important;
+      background-repeat: no-repeat !important;
```
uncenter: This works, but I think we are missing a better solution. Look at where the original Duckduckgo styles set `background-image` — they do it on `.header__logo-icon`, whereas we are setting it here on `.header__logo`/`.js-logo-ddg`. The element we are setting the background image on is the parent of `.header__logo-icon`, and so we don't inherit the `background-repeat: no-repeat;` line that Duckduckgo's default styles already give us!

uncenter (dismissed):

SakethKanchi:
Yeah, good call, thanks. Moved the themed image onto `.header__logo-icon` / `.js-logo-icon` as well and dropped my `!important` repeat line, since that icon already has `no-repeat` from DDG.

I left the parent `.header__logo` in there too because on desktop the icon is `display: none` and DDG paints the parent. On mobile it's the opposite, and DDG clears the parent with `background: none`, so I re-blank it under `&.is-mobile` to match. Without that the parent still tiles.

WalkQuackBack (approved):
Feedback was addressed, LGTM

uncenter (approved):

GenShibe (approved):

PR #2268, opened 2026-07-02
feat(gtfobins): init

ELHart05:
**Website**: GTFOBins

**URL**: https://gtfobins.org

**Description**:

GTFOBins is a curated list of Unix binaries that can be abused to bypass local security restrictions. It is the kind of reference a lot of us keep open during CTFs and pentests. The site is small and static, and it already drives its look from a handful of CSS variables, so it themes cleanly.

**Before** (the site's own theme):

before

**After** (all four flavors, default accent mauve):

| Latte | Frappé |
| :---: | :---: |
| latte | frappe |

| Macchiato | Mocha |
| :---: | :---: |
| macchiato | mocha |

## 🗒 Checklist 🗒

---

- [x] I have read and followed Catppuccin's [userstyle submission guidelines](https://userstyles.catppuccin.com/contributing/creating-userstyles/).
- [x] I have made a new directory underneath `/styles/<name-of-website>`.
  - [x] I have named the userstyle `catppuccin.user.less` within the new directory.
  - [x] I have followed the [template](https://github.com/catppuccin/userstyles/blob/main/template/catppuccin.user.less) and kept the preprocessor as [LESS](https://lesscss.org/#overview).
- [x] I have updated the [`userstyles.yml`](https://github.com/catppuccin/userstyles/blob/main/scripts/userstyles.yml) file with information about the new userstyle.

## 💬 Comments 💬

Almost the whole site is painted from six CSS variables (`--foreground`, `--background`, `--faint`, `--lighter`, `--marked`, `--hover`), so remapping those onto the palette covers most of it in one pass. The rest is the signature red (`#bf0707`), hardcoded in a few spots (links, tag borders, fieldset legends, the anchored-heading marker), which move onto the accent.

I kept the command examples in the plain text color instead of the accent, since they are the main content and accent on the surface color drops below AA contrast in Latte and Frappé. Matched filter tags get an accent tint with a text-color label so they stay readable, and the selection color goes back to `@text` since the site otherwise flips it to the background and it disappears. The logo is left as is, since it is a flat raster with a white background baked in.

Two notes for review: the site is served from gtfobins.org (the github.io URL 301-redirects there), so that is the `@-moz-document` target, and it sets `data-theme` on `<html>` itself, so the style keys off that attribute. Tested live in all four flavors, and `deno fmt` and `deno task lint gtfobins` both pass.


ELHart05:
@WalkQuackBack Any review/comment regarding the PR?, thanks

PR #2294, opened 2026-07-21
feat: add devcontainer config

WalkQuackBack:
Adds a .devcontainer configuration to be used in GitHub Codespaces and VSCode containers.

References:
- Development containers - https://containers.dev/
- [Developing inside a Container](https://code.visualstudio.com/docs/devcontainers/containers)
- [Introduction to dev containers - GitHub Docs](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,3 @@
+FROM denoland/deno:bin-2.9.3 AS deno
+FROM mcr.microsoft.com/devcontainers/typescript-node:24
+COPY --from=deno /deno /usr/local/bin/deno
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,28 @@
+{
+	"name": "Deno",
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// Install git-lfs as it is required to push to the repo
+	"postCreateCommand": "sudo apt update && sudo apt install -y git-lfs && git lfs install",
+
+	// Configure tool-specific properties.
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"justjavac.vscode-deno-extensionpack"
+			]
+		}
+	}
+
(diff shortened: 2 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// Install git-lfs as it is required to push to the repo
+	"postCreateCommand": "sudo apt update && sudo apt install -y git-lfs && git lfs install",
```
uncenter, line 15: Is this specific to our setup? I would think we can avoid this now - we should have removed all the git lfs assets already, maybe we forgot to remove the plumbing?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+
+	// Configure tool-specific properties.
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"justjavac.vscode-deno-extensionpack"
+			]
+		}
```
uncenter, line 23: Can this not just pull from the existing https://github.com/catppuccin/userstyles/blob/main/.vscode/extensions.json?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM denoland/deno:bin-2.9.3 AS deno
```
uncenter, line 1: I would love to avoid having to pin another version of Deno here hmmm.

.devcontainer/devcontainer.json
WalkQuackBack: It's still enabled for the GitHub repository, so any attempts to push changes will error out as it's not installed.

.devcontainer/devcontainer.json
WalkQuackBack: Not sure, it's just what the generator put there when setting up Deno.

.devcontainer/devcontainer.json
uncenter: Shoot. Let me open a tracking issue.

.devcontainer/devcontainer.json
WalkQuackBack: Commented this part out.

.devcontainer/Dockerfile
WalkQuackBack: Any ideas for how to couple this with the root project version?

PR #2295, opened 2026-07-21
chore(pnpm): migrate to 11.15.1 and pin version in package.json

WalkQuackBack:
Migrates pnpm to 11.15.1 and pins the version in package.json.

The version mismatch between CI and local enviornments has been creating issues with working with the documentation project, as in [pnpm 11.x](https://pnpm.io/blog/releases/11.0) onlyBuiltDependencies and related parameters were dropped and replaced with allowBuilds.

In local enviornments, often times the latest version of pnpm would be installed on the local enviornment, which mismatched with the CI, which was locked to a 10.x version. By defining an engines and packageManager field, it avoids mismatch and enforces using a recent enough version of pnpm for the new configuration options.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -3,6 +3,10 @@
   "private": true,
   "type": "module",
   "version": "0.0.1",
+  "packageManager": "pnpm@11.15.1",
+  "engines": {
+    "pnpm": ">=11.8.0"
+  },
   "scripts": {
     "dev": "astro dev",
     "start": "astro dev",
--- a/docs/package.json
+++ b/docs/package.json
@@ -3,9 +3,12 @@
   "private": true,
   "type": "module",
   "version": "0.0.1",
-  "packageManager": "pnpm@11.15.1",
-  "engines": {
-    "pnpm": ">=11.8.0"
+  "devEngines": {
+    "packageManager": {
+      "name": "pnpm",
+      "version": ">=11.8.0 <12.0.0",
+      "onFail": "error"
+    }
   },
   "scripts": {
     "dev": "astro dev",
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,6 @@ jobs:
         uses: withastro/action@a4407e937037e68022f04ae1c068d3634f08bc8d # v4
         with:
           path: ./docs
-          package-manager: pnpm@10.14.0
 
   deploy:
     needs: build
--- a/docs/pnpm-workspace.yaml
+++ b/docs/pnpm-workspace.yaml
@@ -1,3 +1,3 @@
-onlyBuiltDependencies:
-  - esbuild
-  - sharp
+allowBuilds:
+  esbuild: true
+  sharp: true


docs/package.json
```diff
@@ -3,6 +3,10 @@
   "private": true,
   "type": "module",
   "version": "0.0.1",
+  "packageManager": "pnpm@11.15.1",
+  "engines": {
+    "pnpm": ">=11.8.0"
+  },
```
uncenter, line 9: I think these two are supposed to be replaced with `devEngines.pm`.

uncenter: E.g. https://github.com/uncenter/eleventy-plugin-icons/blob/5cf3be7dc814fa8d6f1d8b189e90d118677ec728/package.json#L53-L59.

WalkQuackBack: Yes, that's right! Sorry for the late reply, I didn't see this.

WalkQuackBack:
```suggestion
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=11.8.0 <12.0.0",
      "onFail": "error"
    }
  },
```

How's this look?

WalkQuackBack: Was thinking of potentially bumping up the min version for security reasons, we can determine one to settle on.

uncenter: That snippet change looks good, aside from the indentation missing. Happy to bump to anything v11 but not v12 yet?

WalkQuackBack: pnpm v12 is still in release candidate, so we could consider supporting it after it releases stable.

docs/package.json
```diff
@@ -3,6 +3,13 @@
   "private": true,
   "type": "module",
   "version": "0.0.1",
+  "devEngines": {
+    "packageManager": {
+      "name": "pnpm",
+      "version": ">=11.8.0 <12.0.0",
```
uncenter, line 9: Is there a specific reason to go with 11.8.0 instead of e.g. the latest 11.22.0?

WalkQuackBack: Nothing in particular, I think it was just the version on my personal devenv when I drafted this. If we bump the version to 11.22.0, it might be a good idea to change the onFail value to download so it will download the pinned version of PNPM instead of outright failing.

WalkQuackBack: @uncenter PNPM 12 is now stable. We can switch to it; from my brief look there is no breaking changes and it should be a drop-in replacement.

docs/package.json
```diff
@@ -3,6 +3,13 @@
...
+      "onFail": "error"
```
WalkQuackBack, line 10:
```suggestion
      "onFail": "download"
```

docs/package.json
WalkQuackBack, line 9:
```suggestion
      "version": ">=12.0.0 <12.5.1",
```

uncenter: Sounds good!

uncenter: Should we make it a version range? We should match to some degree whatever the Nix flake uses, for example.

ParasailNumerous: > Should we make it a version range? We should match to some degree whatever the Nix flake uses, for example.

Doesn't look like the docs/ has a Nix flake. I think targeting the latest version is okay, since it will download the latest version if on an older version.

uncenter: I kind of prefer fail over error tbh, but thats nitpicking. Kind of annoying if a bunch of random pnpm versions are downloaded on my system despite my system having a perfectly fine similar v12ish version of pnpm from my system flake.

uncenter: In summary, let's use a caret or a range to say at least v12 or v12.3 or something more flexible.

WalkQuackBack: > In summary, let's use a caret or a range to say at least v12 or v12.3 or something more flexible.

Something like how it was edited? I don't have an opinion on this.

uncenter: Sure that looks good.

PR #2300, opened 2026-07-27
feat(crates.io): rewrite

### LINKED ISSUE #1707: crates.io: unthemed elements

lbl:crates.io

### Describe your problem.

It seems like the CSS variables got changed again, some elements are unthemed.

### Attach screenshots.

Image
Image

Firefox v137.0.1

### Any additional comments?

_No response_

nuexq:
## 🗒 Checklist 🗒

## 🔧 What does this fix? 🔧
rewriting crates.io userstyle following the new site refactor.

## 🔍 Reproduction Steps 🔍
image
image


uncenter:
It looks like there are some unthemed areas on the Code Viewer component (https://crates.io/crates/serde/1.0.229/code/src/lib.rs), on the Security tab (https://crates.io/crates/serde/security), and on some tooltips like those on the Versions tab (https://crates.io/crates/serde/versions, see image):

CleanShot 2026-07-29 at 17 00 24@2x

Also in the API Tokens page (https://crates.io/settings/tokens), there are some hover states of the sidebar menu links that are unthemed and the background of live API token cards are unthemed.

CleanShot 2026-07-29 at 17 03 42@2x

The header in light mode also looks a little strange.

nuexq:
I fixed all the problems you mentioned and found more along the way, and about light mode header, I'm not sure which color to choose as their original color was green color that plays a role as secondary accent for them, and I picked the closest color to the dark mode version, also `@base` and `@mantle` for backgrounds doesn't seem to work well in light mode

styles/crates.io/catppuccin.user.less
```diff
@@ -66,17 +64,41 @@
         --bg-color-top-dark: @red;
         --bg-color-bottom-dark: @red;
       }
-
+      :focus-visible {
+        outline-color: @accent;
+        outline-style: solid;
+      }
```
uncenter: I would move this style (which is very general, as a focus state for everything) to above below the #lib imports instead of being down here.

styles/crates.io/catppuccin.user.less
```diff
@@ -45,215 +29,352 @@
...
+      .input-lg {
+        background-color: @mantle;
+      }
 
-      &:hover {
-        background-color: darken(@accent, 5%);
+      & when (@flavor = latte) {
+        background-color: @text;
       }
```
uncenter: Should we set `--header-bg-color` instead of using this?

uncenter:
CleanShot 2026-08-07 at 14 31 53@2x

These text elements are unthemed on the home page header. Looks like they are `.header .index-link, .header .hero-title`.

uncenter:
It looks like security advisories are unthemed, like that on https://crates.io/crates/hashbrown/security.

uncenter:
The no pending invitations text/card under https://crates.io/me/pending-invites is unthemed.

nuexq:
> CleanShot 2026-08-07 at 14 31 53@2x
> These text elements are unthemed on the home page header. Looks like they are `.header .index-link, .header .hero-title`.

I can't reproduce this state on my end, is there a specific way to trigger it (the 2 badges)?

nuexq:
and the background color of the code preview in https://crates.io/crates/catppuccin/2.8.0/code is `@surface0`, won't it be better to have `@base` as its background color?
| `background-color: @base` | `background-color: @surface0` |
| ------------------------- | ----------------------------- |
|  |  |

uncenter:
> I can't reproduce this state on my end, is there a specific way to trigger it (the 2 badges)?

The red numbered badges are annotations I added with my screen capture tool, highlighting the two instances of unstyled text that should be `@text` ("crates.io" and "The Rust community's crate registry").

uncenter:
> and the background color of the code preview in [crates.io/crates/catppuccin/2.8.0/code](https://crates.io/crates/catppuccin/2.8.0/code) is `@surface0`, won't it be better to have `@base` as its background color?

Go for it! Looks good to me as well.

styles/crates.io/catppuccin.user.less
```diff
@@ -361,21 +416,31 @@
...
+
       .icon {
         color: @subtext0;
       }
     }
+
+    /* Owner Invites */
+    .list:has(p) {
+      background-color: @base;
+    }
```
uncenter, line 444: Why not just apply the background color to `.list` generally? Where and how else do we theme the `.list` component?

nuexq: doing that will cause a lot of problems on other list, like:
image
image

and I didn't find a better way to select the one in Owner Invites than this.

nuexq: a `div.list` selector also would work rather than `.list:has(p)`, as other list are mostly `ul li` and some `ol li`

uncenter: That's sounds better to me!

uncenter:
The header and footer are looking pretty borked if you set the dark theme to latte while in website dark mode, fyi.

nuexq:
do you mean these?
image
image
image

nuexq:
for the green shadows, I don't know of a good way to fix it rather than disabling shadows from everything, or manually remove shadows from the affected elements? and those shadows presented only when I set dark theme to latte. I didn't see other problems from doing that?

scarcekoi:
> for the green shadows, I don't know of a good way to fix it rather than disabling shadows from everything, or manually remove shadows from the affected elements? and those shadows presented only when I set dark theme to latte. I didn't see other problems from doing that?

The shadows are set on crates.io's `:root` like so:
```css
--main-shadow-color: var(--lightningcss-light,var(--green900))var(--lightningcss-dark,#131612);
```

nuexq:
Thanks, didn't notice it. However, shadows are still hardcoded in some other places, such as the input, and I did style them manually to use `--main-shadow-color`.

PR #2301, opened 2026-07-27
feat(scripts): improve library module dev server

### LINKED ISSUE #2293: Document and improve usability of library module dev server
See #1949 for initial tracking.

JosephDemarest:
Closes #2293

## Summary

- serve the selected userstyle at a predictable loopback URL while rewriting hosted library imports to the local server
- add validated CLI arguments, a custom port option, helpful output, and clean shutdown without a temporary userstyle file
- document the library-module development workflow and add focused unit and integration coverage

## Validation

- `deno task test` — 13 passed
- `deno task lint`
- `deno check scripts/**/*.ts`
- `deno fmt --check scripts/serve/main.ts scripts/serve/main.test.ts docs/src/content/docs/contributing/library-modules.md`
- `pnpm build` in `docs` — 18 pages built and all internal links valid
- LSP diagnostics — no findings in the three changed files

### DIFF
--- a/docs/src/content/docs/contributing/library-modules.md
+++ b/docs/src/content/docs/contributing/library-modules.md
@@ -3,4 +3,28 @@ title: Library Modules
 description: Reference for library modules.
 ---
 
-Coming soon.
+Library modules live in the repository's [`lib`](https://github.com/catppuccin/userstyles/tree/main/lib) directory and are hosted at `https://userstyles.catppuccin.com/lib/`. Userstyles should always import the hosted URL so that published styles work outside a local checkout.
+
+## Developing locally
+
+The development server lets a userstyle use your local library files without changing its checked-in imports. It serves a transformed copy of the selected userstyle on the loopback interface and adds a library checksum to each import so that Stylus notices library changes.
+
+1. [Install Deno](https://docs.deno.com/runtime/getting_started/installation/) and clone this repository.
+2. From the repository root, start the server with a userstyle slug or path:
+
+   ```sh
+   deno task serve github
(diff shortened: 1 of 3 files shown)


uncenter:
This pull request appears to be AI generated, please disclose this up front in the future.

I originally refrained from serving the userstyles from a server because I didn't think Stylus would be able to live reload changes. Have you confirmed that Stylus will live reload when the file is changed locally in this new setup?

JosephDemarest:
Yep, AI was used here. I used Fable 5 with max thinking for planning and Sol 5.6 at ultra for execution. I should've disclosed that up front, sorry. I'll do that going forward.

I reviewed the changes, checked the contribution docs and repo requirements, and verified the test results and gates.

I also tested live reload on the current PR with Stylus 2.4.8 on both Chromium and Firefox. With live reload enabled and the installer tab kept open, changes to both the userstyle and library files were picked up and recompiled without warnings. So yes, it works.

JosephDemarest:
Rebased onto current main and moved the development instructions into the new library-modules overview, keeping the upstream documentation changes.

The versioned library layout also needed two compatibility fixes: checksum calculation now includes nested files, and imports inside served library files are rewritten to loopback too, so the legacy lib.less shim does not fetch the hosted standard library. Added regression coverage for both and a real filesystem/HTTP/LESS reload test.

All 16 tests, lint, typecheck and formatting pass. The docs build produces 19 pages with valid internal links. The browser-native Stylus checks from July 27 were not repeated for this refresh; the new verification covers the server and LESS import/recompile path.

PR #2325, opened 2026-09-18
fix(youtube): "new" video thumbnail badge

fe50c:
## 🗒 Checklist 🗒

## 🔧 What does this fix? 🔧

Fixes unthemed "New" badge on some video thumbnails (this may be an A/B tested feature)

## 🔍 Reproduction Steps 🔍

On the youtube.com/watch?v=(video) page, notice that some suggested videos in the right-hand column have "New" badges in the top-left corner of the thumbnail.

### Before
image

*Note the upper left corner*
### After
image

*Note the upper left corner*


styles/youtube/catppuccin.user.less
```diff
@@ -849,6 +849,12 @@
         background: @accent;
       }
     }
+    
+    // "New" indicator
+    .ytBadgeShapeThumbnailSpecial.ytBadgeShapeModern {
+      background: @surface0;
+      color: @text;
+    }
```
uncenter: Instead of inserting this at this point in the file, we should see if we can instead update current instances or sections of code relating to `.ytBadgeShapeThumbnail` - I would try searching for that in this file and seeing what comes up!

fe50c:
That would be a good idea yeah 😭. Thanks for the review despite my stupidity, PR should be fine now, but I haven't been able to find a CC video to double-check. NEW badges work though

